Issue #457, opened 2024-04-17
Add Filters to Newsfeed


chao-mu:
The main use case is a way for people to see only game annotations.

However, listing out every single task would be excessive.

To start, support filtering by game annotations and all the categories (tactics, middlegames+strategy, endgame, etc).

Issue #346, opened 2023-10-27
Newsfeed reaction icon hides blue lets go emote


jalpp:
![Screen Shot 2023-10-27 at 8 54 16 AM](https://github.com/jackstenglein/chess-dojo-scheduler/assets/92553013/ad02da4d-e8c9-484f-9154-eaa98e942ee5)
![Screen Shot 2023-10-27 at 8 56 13 AM](https://github.com/jackstenglein/chess-dojo-scheduler/assets/92553013/818577d6-0ff7-4128-8d05-af3c3a027730)

can see that icon covers up because of dark blue, maybe change it contrast colour?

PR #465, opened 2024-04-22
Feature/457 newsfeed filters cherry pick

chao-mu:
Add category based newsfeed filters

### DIFF
--- a/frontend/src/newsfeed/list/NewsfeedList.tsx
+++ b/frontend/src/newsfeed/list/NewsfeedList.tsx
@@ -10,6 +10,37 @@ import NewsfeedItem from '../detail/NewsfeedItem';
 import LoadMoreButton from './LoadMoreButton';
 import MultipleSelectChip from './MultipleSelectChip';
 
+type FilterMap = Record<string, (entry: TimelineEntry) => boolean>
+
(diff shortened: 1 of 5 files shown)


chao-mu:
References  #457

frontend/src/newsfeed/list/NewsfeedList.tsx
```diff
@@ -152,6 +193,28 @@ const NewsfeedList: React.FC<NewsfeedListProps> = ({
...
+            finalFilters = proposedFilters.filter(
+                (filter) => filter != AllCategoriesFilterName,
+            );
+        }
+
+        setFilters(finalFilters);
+    };
+
+    const allEntries = data?.entries ?? [];
+    const shownEntries = allEntries.filter((entry) =>
```
jackstenglein, line 214: Can we check that showAdditionalFilters is true before performing this operation? If it's false, then shownEntries will always equal allEntries and there isn't any reason to loop through all the entries.

chao-mu:
Problems addressed.

Added an eslint plugin that supports autofix in more cases with loose vs. strict equality. eslint doesn't fix all the cases out of the box in order to support legacy code.

chao-mu:
Fixes #346
Fixes #457

PR #470, opened 2024-04-25
467 enable unauthenticated users to view games

chao-mu:
Allows unauthenticated access to GamePage. Adds a readonly flag to comments that is set if the user is not authenticated.


chao-mu:
fixes #470

PR #484, opened 2024-04-30
refactor(readme)

jalpp:
doing house cleaning, fixing and making readme more dev friendly


frontend/README.md
chao-mu, line 1: npm run dev
npm run start
npm start 

Are all the same command, FYI 

Perhaps worth grouping them or only mentioning npm run dev and delete the other references. Start was supported for backwards compatibility

jalpp: that should fix it

jalpp:
oh lol I closed by mistake, its fine you can re-open if you want

PR #507, opened 2024-05-14
Origin/481 display linter errors in terminal

chao-mu:
Okay, here's the draft of adding rules. Decided to break out into a separate PR.


chao-mu:
Okay, I did a report of what rules get triggered, what the autofix diff would look like, and what the rules would trigger after the autofix. I also included the config and temporary overrides config for reference

https://lint-example.vercel.app/v1

PR #515, opened 2024-05-29
514 feedback and on game import redux

chao-mu:
Still working on test regressions. Might have spotted something introduced in previous release that the preflight makes more obvious. The tests also don't seem to be test user agnostic anymore (was using my own test user by necessity previously).


frontend/src/games/view/GamePage.tsx
```diff
@@ -49,6 +71,61 @@ const GamePage = () => {
...
+        for (const [name, value] of Object.entries(headerMap)) {
+            if (value) {
+                chess.setHeader(name, value);
+            }
+        }
+
+        const update: UpdateGameRequest = {
+            headers,
+            unlisted: true,
+            orientation: 'white',
```
jackstenglein, line 100: Maybe we should add the orientation to this first load popup?

frontend/src/games/view/GamePage.tsx
```diff
@@ -25,8 +35,15 @@ const GamePage = () => {
...
     const request = useRequest<Game>();
     const featureRequest = useRequest();
+    const updateRequest = useRequest();
     const { cohort, id } = useParams();
     const user = useAuth().user;
+    const [searchParams, setSearchParams] = useSearchParams({
+        firstLoad: 'false',
+    });
+    const firstLoad = searchParams.get('firstLoad') === 'true';
+    const [game, setGame] = useState<Game>();
```
jackstenglein, line 45: Why are we adding another state variable instead of just using `request.data`?

frontend/src/board/pgn/boardTools/underboard/settings/GameSettings.tsx
```diff
@@ -162,20 +233,31 @@ const SaveGameButton = ({
...
+            unlisted: unlisted,
+            publish: isPublishing,
             timelineId: game.timelineId,
         };
 
-        if (headers) {
-            update.headers = headers;
+        if (newHeaders) {
+            for (const [name, value] of Object.entries(newHeaders)) {
+                chess?.setHeader(name.charAt(0).toUpperCase() + name.slice(1), value);
```
jackstenglein, line 257: I'd prefer to just change the field names in `GameHeader` to `White`, `Black`, etc instead of doing this string manipulation

frontend/src/board/pgn/boardTools/underboard/settings/GameSettings.tsx
```diff
@@ -38,12 +42,36 @@ interface GameSettingsProps {
...
 const GameSettings: React.FC<GameSettingsProps> = ({ game, onSaveGame }) => {
     const isFreeTier = useFreeTier();
-    const [visibility, setVisibility] = useState(game.unlisted ? 'unlisted' : 'public');
+    const [visibility, setVisibility] = useState(
+        game.unlisted ?? true ? 'unlisted' : 'published',
+    );
     const [orientation, setOrientation] = useState<BoardOrientation>(
         game.orientation ?? 'white',
     );
+    const [dirty, setDirty] = useState(false);
```
jackstenglein, line 51: dirty should be calculated from the state, not a state variable itself. What if I change the orientation and then immediately change it back? The current code will still show as dirty, even though it actually isn't

backend/pgnService/game/update.ts
```diff
@@ -144,60 +138,83 @@ function getRequest(event: APIGatewayProxyEventV2): UpdateGameRequest {
...
         update.white = game.white;
         update.black = game.black;
         update.date = game.date;
-        update.headers = game.headers;
         update.pgn = game.pgn;
+        update.headers = game.headers;
+
+        result = game.headers['Result'] ?? result;
+
+        if (isMissingData({ ...update, result }) && !request.unlisted) {
```
jackstenglein, line 175: This check should probably be against `update.unlisted` instead of `request.unlisted`. What if I passed both `publish: true` and `unlisted: true`? My reading of this code is that it would allow me to publish with missing data.

jackstenglein (changes requested):

backend/pgnService/game/update.ts
```diff
@@ -119,13 +105,21 @@ function getRequest(event: APIGatewayProxyEventV2): UpdateGameRequest {
...
+        if (unlisted === undefined) {
+            throw new ApiError({
+                statusCode: 400,
+                publicMessage: 'Invalid request: unlisted is required',
             });
         }
 
         return {
             ...request,
+            unlisted,
```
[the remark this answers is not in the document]
chao-mu: This is a type checker thing. I changed to Partial as there was no more transpiler enforced type safety that fields existed.

frontend/src/board/pgn/boardTools/boardButtons/StatusIcon.tsx
```diff
@@ -49,6 +49,7 @@ const StatusIcon: React.FC<StatusIconProps> = ({ game }) => {
             api.updateGame(cohort, id, {
                 type: GameSubmissionType.Manual,
                 pgnText,
+                unlisted: game.unlisted ?? true,
```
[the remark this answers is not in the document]
chao-mu: This was an understanding of the existing code for me. I thought unlisted being undefined meant it was still unlisted. I am unsure why I had this Misconception. Thank you!

frontend/src/games/view/GamePage.tsx
chao-mu: I thought similar.

frontend/src/board/pgn/boardTools/underboard/settings/GameSettings.tsx
chao-mu: It was originally that way and I dislike the shortcoming you mentioned, but the upstream dependencies were not keeping in sync and I was out of time. I will revisit to see if I can find a workaround or better articulate the problem.

chao-mu: essentially updating chess headers (form useChess) was not keeping in sync with game (I think from useGame), both firmly outside of the code I was modifying so I declared "here be dragons" and moved on. I'll look back into it in case there was something I was missing.

backend/pgnService/game/update.ts
chao-mu: Since request.unlisted was (changed now) always required, request.unlisted was more relevant than update.unlisted.  I also mistakenly thought undefined unlisted meant unlisted (i.e. games defaulted to unlisted).

frontend/src/board/pgn/boardTools/underboard/settings/GameSettings.tsx
chao-mu: :+1: I agree what I have here is a bit excessive. Thank you.

frontend/src/games/view/GamePage.tsx
chao-mu: Thanks!

chao-mu:
To return err is human,  to if err != nil divine? Thank you for the feedback.

chao-mu:
There's one more issue I need to figure out with date validation and the preflight. Really weird state happened.

frontend/src/api/gameApi.ts
```diff
@@ -220,7 +220,7 @@ export interface UpdateGameRequest extends CreateGameRequest {
     timelineId?: string;
     orientation?: BoardOrientation;
     publish?: boolean;
```
jackstenglein, line 222: Don't forget to remove `publish` here too

frontend/src/board/pgn/boardTools/underboard/settings/GameSettings.tsx
```diff
@@ -43,32 +43,26 @@ interface GameSettingsProps {
 const GameSettings: React.FC<GameSettingsProps> = ({ game, onSaveGame }) => {
     const isFreeTier = useFreeTier();
     const [visibility, setVisibility] = useState(
-        game.unlisted ?? true ? 'unlisted' : 'published',
+        game.unlisted ?? false ? 'unlisted' : 'published',
```
jackstenglein, line 46: Actually we probably don't even need `?? false` here at all

frontend/src/board/pgn/boardTools/underboard/settings/GameSettings.tsx
```diff
@@ -43,32 +43,26 @@ interface GameSettingsProps {
...
+    const dirty =
+        headersChanged ||
+        orientation !== game.orientation ||
+        (game.unlisted ?? false) !== unlisted;
 
-    const onChangeVisibility = (visibility: string) => {
-        setDirty(true);
-        setVisibility(visibility);
-    };
+    const navigate = useNavigate();
```
jackstenglein, line 63: Can you move this up? I like to keep all the hooks together at the top. It helps prevent accidentally returning early before a hook call

jackstenglein (changes requested):

backend/pgnService/game/update.ts
```diff
@@ -144,60 +129,76 @@ function getRequest(event: APIGatewayProxyEventV2): UpdateGameRequest {
...
-async function getGameUpdate(
-    request: UpdateGameRequest,
-): Promise<[GameUpdate | null, GameImportHeaders | null]> {
+async function getGameUpdate(request: UpdateGameRequest): Promise<GameUpdate> {
     const update: GameUpdate = {
         updatedAt: new Date().toISOString(),
+        unlisted: request.unlisted,
     };
 
-    if (request.orientation) {
```
jackstenglein, line 154: This shouldn't have been removed, right?

frontend/src/board/pgn/boardTools/underboard/settings/GameSettings.tsx
chao-mu: Yes, I agree about that best practice. This was a mistake. There exist lint rules to enforce this, by the way.

chao-mu:
This should all be resolved, with the exception of the testing issues mentioned in PM. For some reason the latest commits to address your feedback are grouped under "20 hours ago", but they exist.

backend/pgnService/game/update.ts
```diff
@@ -144,60 +129,81 @@ function getRequest(event: APIGatewayProxyEventV2): UpdateGameRequest {
...
+            update.publishedAt = new Date().toISOString();
+            update.timelineId = `${update.publishedAt?.split('T')[0]}_${uuidv4()}`;
+        }
+    }
+
+    if (request.orientation) {
+        update.orientation = request.orientation;
     }
 
+    let result: string | undefined = request.headers?.result;
```
jackstenglein, line 153: Can we move `result` inside the if since it's not used outside it?

backend/pgnService/game/update.ts
```diff
@@ -144,60 +129,81 @@ function getRequest(event: APIGatewayProxyEventV2): UpdateGameRequest {
...
 
-    return [update, null];
+    return update;
 }
 
 /**
- * Returns the import headers for the Game, if it is missing data. Otherwise returns null.
- * @param game The game to get the import headers for.
- * @returns The import headers, if the game is missing data. Null otherwise.
+ * Strip name header
```
jackstenglein, line 179: Can this comment be a bit more descriptive on what it means to strip the name header?

frontend/cypress/e2e/games/import/importOnlineGame.cy.ts
```diff
@@ -101,14 +105,23 @@ describe('Import Games Page - Import Online Games', () => {
...
         verifyGame({
             lastMove: 'Nxb6',
         });
         deleteCurrentGame();
     });
 
+    it('submits from Chess.com annotations URL (type B)', () => {
+        importUrl(testUrls.chesscomAnalysisB);
+        // This particular analysis is missing headers
+        verifyGame({});
```
jackstenglein, line 121: Shouldn't we still verify the last move? Also, it looks like the analysis does have headers? When I look at it on chesscom, white is named Test1 and black is named Test2. Is that not getting carried over during the import?

backend/pgnService/game/update.ts
chao-mu: Whoops. Looks like old code.

frontend/cypress/e2e/games/import/importOnlineGame.cy.ts
chao-mu: My own copy pasta from Analysis A. I was stuck debugging the weird testing issues I mentioned that I never resolved - hadn't gotten to further refining the new test. I am going to look back into it.

jackstenglein (approved):

PR #558, opened 2024-06-11
557 feat make tactics tests in profile a link

chao-mu:
Okay. This is good to go if it makes sense.

### DIFF
--- a/frontend/src/exams/view/exam.ts
+++ b/frontend/src/exams/view/exam.ts
@@ -203,6 +203,10 @@ export interface TacticsRatingComponent {
 
     /** A description of the component to be displayed to users. */
     description: string;
+
+    /** Indicates rating was derived as a summary (e.g. average) of 0 or more exams
+     * taken within the dojo **/
+    examSummary?: boolean;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


frontend/src/exams/view/exam.ts
```diff
@@ -203,6 +203,10 @@ export interface TacticsRatingComponent {
 
     /** A description of the component to be displayed to users. */
     description: string;
+
+    /** Indicates rating was derived as a summary (e.g. average) of 0 or more exams
+     * taken within the dojo **/
+    examSummary?: boolean;
```
jackstenglein, line 209: I think it makes more sense to add a `link?: string;` attribute and then link to whatever it points to. In the future, we might want to point to different pages for different components, rather than hardcoding a link to the tests page. Also, this allows us to link directly to `/tests/tactics` or `/tests/checkmate` rather than just to `/tests`

frontend/src/profile/stats/TacticsScoreCard.tsx
```diff
@@ -71,7 +71,11 @@ const TacticsScoreCard: React.FC<TacticsScoreCardProps> = ({ user }) => {
                             <Tooltip title={c.description}>
                                 <Stack alignItems='center'>
                                     <Typography variant='body2' color='text.secondary'>
-                                        {c.name}
+                                        {c.examSummary ? (
+                                            <Link href='/tests'>{c.name}</Link>
```
jackstenglein, line 75: Pass the `component={RouterLink}` prop, where `RouterLink` is `import { Link as RouterLink } from react-router-dom;`

jackstenglein (changes requested):

chao-mu:
(hopefully) all good!

jackstenglein (approved):

PR #624, opened 2024-07-05
Nextjs quickfix

chao-mu:
Unfortunately I am not positive I addressed your issues BUT I was able to prevent our stuff from majorly clashing.

- This is now fully an SSR website. Dynamic routes are now supported, among other things.
- I've fixed all build and linting errors
- Blog now has its own default metadata that matches what it was in the old site
- I've grouped the routes ("blog", "legacy", and then my stuff will introduce "scoreboard").
- I've pushed the  Context/Provider wrappers down into blog and legacy, and DRYing it up with a legacy/Layout.tsx. Some if not all of Layout.tsx may end up in src/app/layout.tsx, but my work directly connects to researching the best approach to this, so I'd like neither of us to have to redo a bunch of things.
- `npm run start:build` now runs `next start` and I've removed the `serve` dependency as it's no longer needed. This was previously waiting on changing NextJS's build options away from the migration state

If you want to test what real world performance might be like, you'll need to do npm run `build:development && npm run start:build`. I have not tried this as of yet.

Note, the purpose of the (legacy) routing group is to just house the catch-all react-router shim. I use the term "legacy" for things that are ear marked to be removed or refactored elsewhere. 

There's `legacy/Layout.tsx` which is a stopgap between our current projects.


chao-mu:
Pretty please, if there are major disagreements with how I refactored things, please reach out because some of the reasoning may be non-obvious, especially since it connects with the code I'm working on AND long term plans for migration that I may not have adequately articulated.

PR #653, opened 2024-07-24
refactor(frontend): move pages to app dir

chao-mu:
I'm not currently working on this, but wanted to get the current state of my work into GitHub in case you (Jack) or someone else wants to pick up the torch before I do.

This is nowhere near done, but wanted to create this PR to track the work.


jackstenglein:
Is the goal of this PR to separate every single page from the SPA and move it to the nextjs app router? I would greatly prefer to do that migration in much smaller incremental chunks instead. For example, starting with some of the public facing unauthenticated pages which would benefit most from the improved SEO metadata/performance.

chao-mu:
This was to track historical work I abandoned in case we wanted to do the "all at once" approach at some point.











---- On Tue, 30 Jul 2024 17:30:43 -0400 Jack Stenglein ***@***.***> wrote ---





Is the goal of this PR to separate every single page from the SPA and move it to the nextjs app router? I would greatly prefer to do that migration in much smaller incremental chunks instead. For example, starting with some of the public facing unauthenticated pages which would benefit most from the improved SEO metadata/performance.

PR #670, opened 2024-08-01
635 three dots for puzzles

chao-mu:


frontend/src/exams/view/exam.ts
```diff
@@ -205,6 +205,9 @@ export interface TacticsRatingComponent {
 
     /** Link relevant to this component */
     link?: string;
+
+    /** Number of tests used to calculate this rating */
+    examsCount?: number;
```
jackstenglein, line 210: Could you change the name of this to `examCount`? Otherwise this looks good

Issue #728, opened 2024-08-24
Training plan sparring position overflow on mobile 


chao-mu:
If you open a training plan task with a position in it on mobile, the position stretches outside the bounds of the screen
So any of the middlegame sparring tasks, endgame sparring, etc will work as examples

chao-mu:
![image](https://github.com/user-attachments/assets/a2d0eb70-0232-42fd-9978-b21ca6dba8a8)

chao-mu:
And on desktop it currently looks like this

![image](https://github.com/user-attachments/assets/08be6dc1-5c06-4226-96c4-d68097e50ce4)

chao-mu:
![image](https://github.com/user-attachments/assets/e77f6865-3c5d-49b4-bf7b-5e1a2c7ddfa0)

chao-mu:
![image](https://github.com/user-attachments/assets/f08c235d-4323-4ba3-a0a7-ad627b6caabf)

chao-mu:
![image](https://github.com/user-attachments/assets/d1c8cbce-3fa0-41e7-b671-6758cd68830b)

PR #729, opened 2024-08-25
fix(#728): Position overflows on requirement review, sparring page, a…

chao-mu:
It's now responsible for the consumer of Position to control width, which makes it easier to be responsive.


jackstenglein:
Looks like it's still slightly cut off on the Import from Custom Position screen:

Screenshot 2024-08-25 at 10 26 19 AM

jackstenglein:
Otherwise, it looks good to me

jackstenglein (approved):

PR #769, opened 2024-09-02
feat(#768): add ability to cancel tutorial

chao-mu:
No longer draft.

### DIFF
--- a/frontend/src/tutorial/Tutorial.tsx
+++ b/frontend/src/tutorial/Tutorial.tsx
@@ -54,6 +54,7 @@ const Tutorial: React.FC<TutorialProps> = ({ name, steps, zIndex }) => {
                         zIndex: zIndex || 100,
                     },
                 }}
+                disableOverlayClose
                 scrollOffset={100}
                 callback={callback}
             />
--- a/frontend/src/tutorial/Tutorial.tsx
+++ b/frontend/src/tutorial/Tutorial.tsx
@@ -1,8 +1,8 @@
 import { useCallback, useEffect, useMemo } from 'react';
 import ReactJoyride, { CallBackProps, Step } from 'react-joyride';
 
-import { useApi } from '../api/Api';
-import { useAuth } from '../auth/Auth';
+import { useApi } from '@/api/Api';
+import { useAuth } from '@/auth/Auth';
 import { useTutorial } from './TutorialContext';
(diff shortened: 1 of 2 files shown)


jackstenglein:
The IconButton has a weird shape while hovering over it: Screenshot 2024-09-02 at 11 28 18 AM

frontend/src/tutorial/Tutorial.tsx
```diff
@@ -58,16 +54,14 @@ const Tutorial: React.FC<TutorialProps> = ({ name, steps, zIndex }) => {
                         zIndex: zIndex || 100,
                     },
                 }}
-                disableCloseOnEsc
-                disableOverlayClose
```
jackstenglein, line 62: Can we leave disableOverlayClose that was somebody doesn't accidentally close it?

frontend/src/tutorial/TutorialTooltip.tsx
```diff
@@ -19,14 +20,24 @@ const TutorialTooltip: React.FC<TooltipRenderProps> = ({
...
             data-cy='tutorial-tooltip'
             sx={{ maxWidth: 'md', width: { xs: '95vw', sm: undefined } }}
             {...tooltipProps}
         >
-            <CardHeader title={step.title} />
+            <CardHeader
+                title={step.title}
+                action={
+                    <IconButton arra-label='close' onClick={closeProps.onClick}>
+                        <Typography component='span' color='Gray'>
```
jackstenglein, line 35: Removing the typography component here and just passing CloseIcon will probably fix the weird shape. You can pass a color directly to the close icon. Also, can we put gray in lowercase?

chao-mu: Sadly, the color prop type does not accept this value - unless my memory fails me, hence the typography. However, the weird shape (if you're talking about the one on hover) is there regardless. I will try again with these regards though, if I can't fix it with the methods you advise, what then?

frontend/src/tutorial/Tutorial.tsx
chao-mu: The first time I used it I kept clicking on the overlay trying to get it to close. It's the trained behavior of many apps that clicking away from the dialog closes it. Maybe this is a question for Post?

jackstenglein (approved):

PR #781, opened 2024-09-07
779 tables not usable on mobile (DRAFT)

chao-mu:
So far this is just the refactor I needed to do to greatly simplify my changes and share across the code base. I tried to hit as minimal components as possible and not do anything premature,  but it did have some consequences that extended past it.

I needed to pull out the GameTable into its own component and as such it triggered needing to do some refactoring to keep things organized.

Something we talked about awhile back was moving shared components into src/components. If a component is page specific, it remains in that current directory, but if multiple pages can and/or do use it, it gets shoved in src/components.

Part of my motivation for this design is that currently our dependency graph has lots of instances of Page -> Child Component -> Different Page, which ends up making NextJS server components tremendously difficult as incompatible dependencies get pulled in, as pretty soon the entire dependency graph has to be included to render a single page.

As we move forward, this segmentation will mean it is easier to state rules like "don't import components from a page's directory from a different page". This will also make for much prettier import paths as we move to NextJS and do some needed decoupling (lots of pure logic we may need server side in tsx files). 

There's lots of cross pollination in the code base, including reusable components I didn't know about, having assumed they were page specific before seeing them in a ton of locations.

Also, maybe I'm alone, but I think a lot about the dependency graph when understanding how an application is structured.

Anyways, here's the refactor.


frontend/src/profile/directories/columns.tsx
jackstenglein, line 1: What is the purpose of this file? Is this being used somewhere?

PR #802, opened 2024-09-24
779 game table improvements

chao-mu:


frontend/src/components/games/list/GameListItem.tsx
```diff
@@ -153,6 +235,38 @@ export function RenderResult(params: GridRenderCellParams) {
     );
 }
 
+function parseTimeControl(
```
jackstenglein, line 238: Parsing a time control is actually more complicated than this, but we can probably use this for the majority of games that are being submitted by users. There's probably some master games where this won't be accurate, but not a big deal

frontend/src/components/games/list/GameListItem.tsx
```diff
@@ -153,6 +235,38 @@ export function RenderResult(params: GridRenderCellParams) {
...
+    return [Math.round(gameTime / 60), increment];
+}
+
+export function getTimeControl({ timeControl }: { timeControl?: string }) {
+    const [gameTime, increment] = parseTimeControl(timeControl);
+
+    return gameTime ? `${gameTime}+${increment ?? 0}` : '';
+}
+
+export function RenderTimeControl({ timeControl }: { timeControl?: string }) {
```
jackstenglein, line 262: Should this return null if timeControl is undefined/empty rather than rendering an empty div/p?

frontend/src/components/games/list/GameTable.tsx
```diff
@@ -18,115 +18,110 @@ import {
...
         valueGetter: (_value, row) =>
             `${row.headers.White} (${row.headers.WhiteElo}) - ${row.headers.Black} (${row.headers.BlackElo})`,
         renderCell: RenderPlayersCell,
-        flex: 1,
-        minWidth: 150,
+        minWidth: 200,
+        flex: 2,
     },
     {
-        field: 'result',
```
jackstenglein, line 49: One side effect of removing this column is that we lose the ability to sort by result, which is a rather big loss

frontend/src/components/ui/CardGrid.tsx
jackstenglein, line 1: Did I miss it somewhere or is this file not used?

frontend/src/components/games/list/GameCard.tsx
jackstenglein, line 1: Did I miss it somewhere or is this file not used?

frontend/src/components/games/list/GameListItem.tsx
chao-mu: Do we have examples of the logic elsewhere? I know there are multiple time increments... The logic here is specifically for Lichess and I'm wondering if I should check that and if it's not Lichess use verbatim?

jackstenglein: You can look at the Clock Usage underboard tab.

Also, a full description of the TimeControl header is here: http://www.saremba.de/chessgml/standards/pgn/pgn-complete.htm#c9.6.1

Again, the vast majority of games will have a simple time control header, so this is pretty low priority

frontend/src/components/games/list/GameTable.tsx
```diff
@@ -91,9 +149,16 @@ export default function GameTable({
     contextMenu,
     limitFreeTier,
     columns,
+    defaultVisibility,
 }: GameTableProps) {
     const freeTierLimited = useFreeTier() && limitFreeTier;
     const { data, request, page, pageSize, rowCount, hasMore, setPage } = pagination;
+    const [columnVisibility, setColumnVisibility] = useState<GridColumnVisibilityModel>({
```
jackstenglein, line 156: If you're wanting to make these settings sticky, you can do it very easily by replacing the useState call with useLocalStorage

frontend/src/components/games/list/GameListItem.tsx
chao-mu: Okay, I didn't mess too much with it, but looking at ClockUsage showed me there was existing functions I could use so I dropped them in.

frontend/src/components/games/list/GameTable.tsx
chao-mu: Was added back in.

frontend/src/components/games/list/GameTable.tsx
chao-mu: Done. Added a namespace since different pages will want different columns visible (e.g. your own games versus other's)

chao-mu:
Ready for final code review.

frontend/src/profile/GamesTab.tsx
```diff
@@ -20,6 +20,7 @@ interface GamesTabProps {
 }
 
 const GamesTab: React.FC<GamesTabProps> = ({ user }) => {
+    const navigate = useNavigate();
```
jackstenglein, line 23: Can you revert these changes related to useNavigate()? It breaks the headers NextJS has to return to enable Stockfish. Also, we're eventually going to stop using react-router-dom and use the NextJS app router instead

chao-mu: What should I do instead? These are existing changes

PR #834, opened 2024-10-02
fix credits for Stockfish

jalpp:
add credits for Lichess/stockfish.js/Stockfish and freechess


frontend/next.config.mjs
jackstenglein, line 1: Can you revert the change to this file please?

jalpp:
Will do

PR #898, opened 2024-11-13
Dep graph linting and docs

chao-mu:
Generate documentation and dependency graphs. Also supports import linting. 

We can eventually use the lint rules to prevent react-router-dom and other breaking imports from being included by NextJS code. We can also use it to detect and reduce circular dependencies. And also use it to gauge impact.

This code was laying around and I wanted to use it to help plan next steps in my save workflow code.


chao-mu:
Ugh. Having issues isolating my old commit.

PR #960, opened 2024-12-03
feat(#920): add options to merge PGN into game

jackstenglein:


chao-mu:
Do you want my code for merging N online games into a provided PGN? It cuts off after a max depth, leaving a comment of what game. It's sort of like a built in player database.

That said being able to restrict the database of moves to just the players would mean what I wrote (It was designed as a repitoir review) is obsolescent.

Issue #918, opened 2024-11-21
Heatmap doesn't update until refreshing after updating activity


jackstenglein:
This happens because the useTimeline hook doesn't cache the response in a provider (for example, like the clubs and events data). If we cache the data and then update it when progress is changed, then that should fix this.

PR #962, opened 2024-12-06
Issue(#918) - Heatmap now updates on activity

clarkevandenhoven:
Previously, the timeline was implemented through a hook `useTimeline` that on page load would create a request object and fetch the current activity status for the heatmap (and other timeline users).

However, when activity was updated, there was no mechanism to signal the `useTimeline` hook to update its information, since each instance of the hook was local the component it was mounted with.

`useTimelime` was therefore changed to `useTimelineContext`, which allows the updateUserProgress call to signal the (now global) timeline context to update the locally stored activity.

Beyond that @jackstenglein I think the timeline context is now a little misplaced, let me know where you think it should go!

### DIFF
--- a/frontend/src/profile/progress/ProgressUpdater.tsx
+++ b/frontend/src/profile/progress/ProgressUpdater.tsx
@@ -27,6 +27,7 @@ import {
     isRequirement,
 } from '../../database/requirement';
 import InputSlider from './InputSlider';
+import { useTimelineContext } from '../../profile/activity/useTimeline';
 
 const NUMBER_REGEX = /^[0-9]*$/;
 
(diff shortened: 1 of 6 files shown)


frontend/src/profile/progress/ProgressUpdater.tsx
```diff
@@ -170,6 +172,8 @@ const ProgressUpdater: React.FC<ProgressUpdaterProps> = ({
                 setHours('');
                 setMinutes('');
                 request.reset();
+                // The timeline needs to know that the update request is done
+                timelineLoadMore();
```
jackstenglein, line 176: I'm not sure that this will do the correct thing if the user has enough timeline activity that `startKey` has a value in the useTimelineContext. If `startKey` has a value, then calling load more will start loading timeline entries from the start key, which means the newly created activity won't be included.

clarkevandenhoven: Resolved by adding a new method `resetRequest` (which also has a clearer name) that also clears the `startKey`, forcing the search to get the newly posted activity.

PR #1054, opened 2025-01-07
Remove Unused Dependencies

chao-mu:
Any of this seem funky? Final if the checks pass.


frontend/package.json
```diff
@@ -118,49 +111,33 @@
...
         "@types/react-resizable": "^3.0.8",
         "@types/uuid": "^10.0.0",
         "@typescript-eslint/eslint-plugin": "^7.12.0",
         "@typescript-eslint/parser": "^7.12.0",
-        "babel-plugin-istanbul": "^7.0.0",
-        "command-line-args": "^6.0.1",
-        "command-line-usage": "^7.0.3",
         "cypress": "^13.17.0",
         "cypress-parallel": "^0.14.0",
-        "dependency-cruiser": "^16.5.0",
```
jackstenglein, line 143: Didn't you just add dependency-cruiser a few PRs ago? Do we no longer need it?

chao-mu: Oh my goodness, good catch.

chao-mu: I did, but you launch it with npx according to the docs, so it's not needed. The important part is the config that was generated. We haven't integrated it fully into our deployment process yet.

Issue #1061, opened 2025-01-10
Add Discord link to private server somewhere visible


hellokostya:
Currently the Discord link is somewhat buried inside a task, but someone may forget where to find it. Can we add a button for it somewhere, maybe even in the ||| three-bars/menu?

hellokostya:
Relatedly, can we add a link to the public server inside the task that says "Join the Discord"

Some folks have already joined the free Discord and then see that task and think they already joined the Discord, so I think we need to make it clear that there's a separate private server for members to join. So maybe have two separate links in that task description:

TRAINING PROGRAM DISCORD (link 1)

PUBLIC DISCORD (link 2)

PR #1081, opened 2025-01-13
fix(faq): fix Discord visibility

jalpp:
solves #1061


jalpp:
@jackstenglein I wasn't able to properly add Discord invite button in the task section, but I think faq section does the work

PR #1107, opened 2025-01-14
feat: add training tips to help stuck/confused dojoers

chao-mu:
A Training Tips button

### DIFF
--- a/frontend/src/components/profile/TrainingTips.tsx
+++ b/frontend/src/components/profile/TrainingTips.tsx
@@ -1,3 +1,4 @@
+import { EventType, trackEvent } from '@/analytics/events';
 import { icons } from '@/style/Icon';
 import {
     Box,
@@ -21,6 +22,11 @@ import { Link } from '../navigation/Link';
 export function TrainingTipsButton() {
     const [showDialog, setShowDialog] = useState(false);
 
+    const onOpen = () => {
+        trackEvent(EventType.OpenProgramTips);
+        setShowDialog(true);
+    };
+
     return (
         <>
             {showDialog && (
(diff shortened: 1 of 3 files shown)


jackstenglein (changes requested):
Go to the profile, open the program tips dialog, click "annotate your games", choose "starting position", then turn on the engine. The engine will fail to load due to COEP error:

Screenshot 2025-01-14 at 5 42 06 PM

frontend/src/components/profile/TrainingTips.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+    return (
+        <Grid2
+            size={{
+                xs: 12,
+                sm: 6,
+            }}
+        >
+            <Card sx={{ height: 1 }}>
+                <CardActionArea sx={{ height: 1 }} onClick={onClick}>
```
jackstenglein, line 144: Use `href` instead of `onClick` and add the `component` parameter. Pass it `Link` from `@/components/navigation/Link`. That should fix the engine load error.

frontend/src/components/profile/TrainingTips.tsx
```diff
@@ -0,0 +1,169 @@
...
+export function TrainingTipsButton() {
+    const [showDialog, setShowDialog] = useState(false);
+
+    return (
+        <>
+            {showDialog && (
+                <TrainingTipsDialog
+                    closeDialog={() => setShowDialog(false)}
+                    open={showDialog}
+                    onClose={() => setShowDialog(false)}
```
jackstenglein, line 30: Why do we have both `closeDialog` and `onClose` with the exact same implementation?

chao-mu:
> Go to the profile, open the program tips dialog, click "annotate your games", choose "starting position", then turn on the engine. The engine will fail to load due to COEP error:
> Screenshot 2025-01-14 at 5 42 06 PM

Either using the Link component fixed it or I am inclined to believe this is coincidence. I was able to reproduce when I started, but not now.

jackstenglein (approved):

PR #1160, opened 2025-01-31
feat(badges): add logic to give out eligible badges

jalpp:


frontend/src/components/profile/info/BadgeHandler.ts
```diff
@@ -0,0 +1,252 @@
...
+ * Gets all possible elgible badges for user
+ * @param user the current user
+ * @returns all possible badges
+ */
+export function getEligibleBadges(user: User): string[][] {
+    const keys = Object.values(BADGE);
+    const overallInfo: string[][] = [];
+    keys.forEach((key) => {
+        const badgeImg = isEligibleBadgeImage(user, key);
+        if (badgeImg != undefined) {
```
jackstenglein, line 246: This can be written as `if (badgeImg)` (although again I would prefer if this was `badgeImage`)

Also, you should basically never use `!=` and `==` in JavaScript. Use `!==` and `===` instead.

frontend/src/components/profile/info/BadgeHandler.ts
```diff
@@ -0,0 +1,252 @@
...
+    return msg;
+}
+
+/**
+ * gets the info and badge image for eligible badge
+ * @param user current user
+ * @param badge the badge type
+ * @returns info and image for the badge
+ */
+function isEligibleBadgeImage(user: User, badge: BADGE): string[] | undefined {
```
jackstenglein, line 222: The function name implies this will return a boolean but it returns a string list instead

frontend/src/components/profile/info/BadgeHandler.ts
```diff
@@ -0,0 +1,252 @@
...
+    const level: number = isEligibleForLimit(user, badge, BADGE_LIMITS[badge]);
+    const info: string[] = [];
+    if (level == -1) {
+        return undefined;
+    }
+
+    info.push(getBadgeImage(level, badge));
+    info.push(BADGE_TITLE[badge] + ' ' + level);
+    info.push(getBadgeMessage(level, badge));
```
jackstenglein, line 231: Instead of arbitrarily pushing these in a random order, you should return an object like 

```
{
  image: getBadgeImage(level, badge),
  title: ...,
  message: ...,
}
```

The way you have it currently requires me to look at this function to know which order the strings are in. That order is easy to forget when calling the function and using the result elsewhere, which makes the rest of the code more error prone. If you return an object, then you will always know which is which.

frontend/src/components/profile/info/BadgeHandler.ts
```diff
@@ -0,0 +1,252 @@
...
+}
+
+/**
+ * gets the level the user is elgible for
+ * @param user the current user
+ * @param reqid the requirement id
+ * @param levels the counting badge levels list
+ * @returns the current level badge user can get
+ */
+function isEligibleForLimit(user: User, reqid: string, levels: number[]): number {
```
jackstenglein, line 169: This function name implies it will return a boolean but instead it returns a number

frontend/src/components/profile/info/BadgeHandler.ts
```diff
@@ -0,0 +1,252 @@
...
+    let maxLevel: number = -1;
+    for (const level of levels) {
+        const elgible = isEligibleForBadge(user, reqid, level);
```
jackstenglein, line 172: This should be spelled `eligible`

frontend/src/components/profile/info/BadgeHandler.ts
```diff
@@ -0,0 +1,252 @@
...
+
+/**
+ * gets the level the user is elgible for
+ * @param user the current user
+ * @param reqid the requirement id
+ * @param levels the counting badge levels list
+ * @returns the current level badge user can get
+ */
+function isEligibleForLimit(user: User, reqid: string, levels: number[]): number {
+    let maxLevel: number = -1;
```
jackstenglein, line 170: The type annotation is unnecessary when you initialize the variable to -1

frontend/src/components/profile/info/BadgeHandler.ts
```diff
@@ -0,0 +1,252 @@
...
+/**
+ * gets the info and badge image for eligible badge
+ * @param user current user
+ * @param badge the badge type
+ * @returns info and image for the badge
+ */
+function isEligibleBadgeImage(user: User, badge: BADGE): string[] | undefined {
+    const level: number = isEligibleForLimit(user, badge, BADGE_LIMITS[badge]);
+    const info: string[] = [];
+    if (level == -1) {
```
jackstenglein, line 225: Don't use `==` in JavaScript. Use `===` instead.

frontend/src/components/profile/info/BadgeHandler.ts
```diff
@@ -0,0 +1,252 @@
...
+}
+
+/**
+ * check if a user is eligible for the badge and given requirement
+ * @param user the current user
+ * @param reqid the requirement ID
+ * @param targetLevel the target level for user to earn the badge
+ * @returns if user is eligible
+ */
+function isEligibleForBadge(user: User, reqid: string, targetLevel: number): boolean {
```
jackstenglein, line 148: In keeping with the above comment, `reqid` should be `requirementId`

frontend/src/components/profile/info/BadgeHandler.ts
```diff
@@ -0,0 +1,252 @@
...
+    const progress = user.progress[reqid];
+    if (!progress) {
+        return false;
+    }
+
+    let totalCount = 0;
+    for (const key in progress.counts) {
```
jackstenglein, line 155: You are calculating this total count every time you call `isEligibleForBadge` in `isEligibleForLimit`. It would be better to calculate this once in `isEligibleForLimit` and then re-use that variable to find the correct limit.

frontend/src/components/profile/info/BadgeCard.tsx
```diff
@@ -1,12 +1,55 @@
...
+} from '@mui/material';
 import Image from 'next/image';
+import { useEffect, useState } from 'react';
 import postmortem2023 from './2023-postmortem.png';
 import postmortem2024 from './2024-postmortem.png';
+import { getEligibleBadges } from './BadgeHandler';
 
 export const BadgeCard = ({ user }: { user: User }) => {
+    const [openDialog, setOpenDialog] = useState(false);
+    const [selectedBadge, setSelectedBadge] = useState<string[] | null>(null);
```
jackstenglein, line 25:
```
const [selectedBadge, setSelectedBadge] = useState<string[]>();
```

This will set `selectedBadge` to `undefined` and does not need an extra type annotation. Prefer to use `undefined` over `null`.

Also, `openDialog` currently just seems like a proxy for `selectedBadge !== undefined`. Is there ever a case where `openDialog` is true but `selectedBadge` is not set? If there is not such a case, remove `openDialog` and just check if selectedBadge is set.

frontend/src/components/profile/info/BadgeHandler.ts
jackstenglein, line 1: This file doesn't have any React components, so it should be in camel case instead of pascal case.

frontend/src/components/profile/info/BadgeCard.tsx
```diff
@@ -1,12 +1,55 @@
...
+
+    const badgeImgs: string[][] = getEligibleBadges(user);
```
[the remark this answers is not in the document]
jackstenglein: Actually, I read through the code some more and realized that this doesn't just contain image URLs. We should call it something like `badgeData` instead.

jackstenglein (changes requested):

frontend/src/components/profile/info/BadgeCard.tsx
```diff
@@ -39,26 +83,88 @@ export const BadgeCard = ({ user }: { user: User }) => {
                 <Tooltip title='View my 2023 postmortem!'>
                     <Image
                         src={postmortem2023}
-                        style={{ height: '40px', width: '40px' }}
                         alt='2023 postmortem'
+                        width={70}
```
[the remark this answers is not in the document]
jalpp: yea that was a typo when I was testing different badge sizes, fixed!

frontend/src/components/profile/info/BadgeHandler.ts
jalpp: I don't really understand this, right now it calculates total count for each level for badge limit levels I think its correct

frontend/src/components/profile/info/BadgeCard.tsx
jalpp: yup I removed openDialog it was just a proxy

frontend/src/components/profile/info/BadgeHandler.ts
jalpp: fixed!

frontend/src/components/profile/info/BadgeCard.tsx
```diff
@@ -1,17 +1,61 @@
...
 import Image from 'next/image';
+import { useEffect, useState } from 'react';
 import postmortem2023 from './2023-postmortem.png';
 import postmortem2024 from './2024-postmortem.png';
+import { Badge, getEligibleBadges, getTacticsChampionBadge } from './badgeHandler';
 
 export const BadgeCard = ({ user }: { user: User }) => {
+    const [selectedBadge, setSelectedBadge] = useState<Badge | undefined>(undefined);
+    const badgeData: Badge[] = getEligibleBadges(user);
+    const [previousBadgeImgs, setPreviousBadgeImgs] = useState<Badge[]>(badgeData);
```
jackstenglein, line 28: `previousBadgeImgs` and `setPreviousBadgeImgs` also need to be renamed

frontend/src/components/profile/info/BadgeCard.tsx
```diff
@@ -1,17 +1,61 @@
...
+        const newBadge = badgeData.find(
+            (badge, index) =>
+                !previousBadgeImgs[index] ||
+                badge.image !== previousBadgeImgs[index].image,
+        );
+        if (newBadge) {
+            setSelectedBadge(newBadge);
+        }
+        setPreviousBadgeImgs([...badgeData]);
+    }, [badgeData]);
```
jackstenglein, line 52: We should only ever be adding badges, so I think you could simplify this a lot with something like:

```
if (previousBadgeImgs.length < badgeData.length) {
    setSelectedBadge(badgeData[previousBadgeImgs.length]);
    setPreviousBadgeImgs(badgeData);
}
```

Note that we also don't need to recreate the array when calling `setPreviousBadgeImgs`. Finally, the useEffect dependency array should be `[badgeData, previousBadgeImgs, setSelectedBadge, setPreviousBadgeImgs]`. Any variables used in the function need to be in the dependency array, or you will probably see incorrect/weird behavior.

frontend/src/components/profile/info/BadgeCard.tsx
```diff
@@ -39,26 +84,155 @@ export const BadgeCard = ({ user }: { user: User }) => {
...
                         alt='2023 postmortem'
+                        width={50}
+                        height={50}
                     />
                 </Tooltip>
             </Link>,
         );
     }
 
+    if (!isProvisional && tacticsRating.overall > minCohort) {
```
jackstenglein, line 96: Can we remove this tactics champion badge for now? This could disappear when someone graduates, which seems like a poor user experience. We'll have to discuss this in the next meeting.

frontend/src/components/profile/info/BadgeCard.tsx
```diff
@@ -39,26 +84,155 @@ export const BadgeCard = ({ user }: { user: User }) => {
...
+                    onClick={() => handleBadgeClick(champion)}
+                    onMouseEnter={(e) => (e.currentTarget.style.transform = 'scale(1.1)')}
+                    onMouseLeave={(e) => (e.currentTarget.style.transform = 'scale(1)')}
+                    className={champion.israre ? 'glow' : ''}
+                />
+            </Tooltip>,
+        );
+    }
+
+    if (badgeData.length !== 0) {
```
jackstenglein, line 122: This `if` is unnecessary. The `for` loop will not run if badgeData is empty.

frontend/src/components/profile/info/BadgeCard.tsx
```diff
@@ -39,26 +84,155 @@ export const BadgeCard = ({ user }: { user: User }) => {
...
+                                : undefined,
+                            borderRadius: '8px',
+                            transition: 'transform 0.2s',
+                        }}
+                        alt={badge.title}
+                        onClick={() => handleBadgeClick(badge)}
+                        onMouseEnter={(e) =>
+                            (e.currentTarget.style.transform = 'scale(1.1)')
+                        }
+                        onMouseLeave={(e) =>
```
jackstenglein, line 143: Rather than defining `onMouseEnter` and `onMouseLeave` event listeners for this, can we add a hover style that sets the transform and do this entirely in CSS?

frontend/src/components/profile/info/BadgeCard.tsx
```diff
@@ -39,26 +84,155 @@ export const BadgeCard = ({ user }: { user: User }) => {
...
+                    </Stack>
+                </CardContent>
+            </Card>
+
+            <Dialog open={selectedBadge !== undefined} onClose={handleCloseDialog}>
+                {selectedBadge && (
+                    <>
+                        <DialogTitle
+                            sx={{
+                                backgroundColor: '#222222',
```
jackstenglein, line 173: I'm not a big fan of defining colors like this. For example, on light mode, it is basically impossible to see the close IconButton outline when you hover over it. I would prefer instead to use the default background colors provided by MUI for the dialog title and dialog content. It keeps all the components and the overall design of the site consistent.

I assume you did this so that the badge glow would be visible? I would rather change the badge glow depending on the color mode of the user. In the theme provider, you could define a new color called `badgeGlow` or something similar and set it differently for light/dark mode and then use that as the badge glow color.

frontend/src/components/profile/info/badgeHandler.ts
```diff
@@ -0,0 +1,312 @@
...
+        (sum, v) => sum + v,
+        0,
+    );
+
+    // make searching in O(1) if user hits max return the max level
+    if (totalCount >= levels[levels.length - 1]) {
+        return levels[levels.length - 1];
+    }
+    // same but the start useful for mate in 3 or badges with 1 max level again O(1) search time
+    if (levels.length === 1 && totalCount >= levels[0]) {
```
jackstenglein, line 188: This `if` is unnecessary. If `levels` has length 1, then the for loop below will only run one time. Also, if `levels` has length 1, then this is equivalent to the `if` condition above.

frontend/src/components/profile/info/badgeHandler.ts
```diff
@@ -0,0 +1,312 @@
...
+        return -1;
+    }
+
+    const totalCount = Object.values(progress.counts || {}).reduce(
+        (sum, v) => sum + v,
+        0,
+    );
+
+    // make searching in O(1) if user hits max return the max level
+    if (totalCount >= levels[levels.length - 1]) {
```
jackstenglein, line 184: I don't think we need this `if`. Our badge levels are short so this won't save much time. Also, since most users will not be at the max level for most badges, this would probably actually be slower than just not having it.

frontend/src/components/profile/info/badgeHandler.ts
```diff
@@ -0,0 +1,312 @@
...
+}
+
+/**
+ * gets the badge title
+ * @param level the current level
+ * @param badge the badge type
+ * @returns the badge title
+ */
+function getBadgeTitle(level: number, badge: BADGE): string {
+    return `${BADGE_TITLE[badge]} ${level}`;
```
jackstenglein, line 242: Can we add a dash or something in between the badge title and level? It looks a bit weird for example as `Classical Games Played 75`

frontend/src/components/profile/info/badgeHandler.ts
```diff
@@ -0,0 +1,312 @@
...
+    title: string;
+    message: string;
+    israre: boolean;
+    rareglowhexcode?: string;
+}
+
+/**
+ * BADGE enum representation
+ */
+export enum BADGE {
```
jackstenglein, line 14: Enums should be in PascalCase, not all caps. Same with the enum properties

frontend/src/components/profile/info/badgeHandler.ts
```diff
@@ -0,0 +1,312 @@
...
+}
+
+/**
+ * check if the user is eligible for badge limit for given requirement
+ * @param user the current user
+ * @param requirementId requirement id
+ * @param levels the badge requirement levels
+ * @returns the badge limit user is eligible for level
+ */
+function getEligibleForLimitLevel(
```
jackstenglein, line 168: Can we call this `getBadgeLevel`?

frontend/src/components/profile/info/badgeHandler.ts
```diff
@@ -0,0 +1,312 @@
+import { User } from '@/database/user';
+
+export interface Badge {
+    image: string;
+    title: string;
+    message: string;
+    israre: boolean;
```
jackstenglein, line 7: `israre` seems useless and is only set when `rareglowhexcode` is set. I would remove it.

frontend/src/components/profile/info/badgeHandler.ts
```diff
@@ -0,0 +1,312 @@
...
+    rareglowhexcode?: string;
```
jackstenglein, line 8: This should be in camelCase. Also, we can remove the notion of rarity and just define it as `glowHexcode`. This allows us to have future badges that glow even if they are not deemed rare and the variable name still makes sense.

frontend/src/components/profile/info/badgeHandler.ts
```diff
@@ -0,0 +1,312 @@
...
+    if (level === -1) {
+        return undefined;
+    }
+
+    const currentBadge: Badge = {
+        image: getBadgeImage(level, badge),
+        title: getBadgeTitle(level, badge),
+        message: getBadgeMessage(level, badge),
+        israre: isRareBadge(level, badge),
+        rareglowhexcode: isRareBadge(level, badge) ? BADGE_RARE_COLOR[badge] : undefined,
```
jackstenglein, line 273: In accordance with removing `isRare` and changing this to `glowHexcode`, I think it would make more sense to have a function like `getBadgeGlow(level, badge)` which returns a hexcode or undefined if the badge should not glow.

frontend/src/components/profile/info/badgeHandler.ts
```diff
@@ -0,0 +1,312 @@
...
+/**
+ * Gets all possible elgible badges for user
+ * @param user the current user
+ * @returns all possible badges
+ */
+export function getEligibleBadges(user: User): Badge[] {
+    const keys = Object.values(BADGE);
+    const overallInfo: Badge[] = [];
+    keys.forEach((key) => {
+        const badgeImage = getEligibleBadgeInfo(user, key);
```
jackstenglein, line 305: This is no longer just an image, so should be named `badge` instead.

jackstenglein (changes requested):

frontend/src/components/profile/info/badgeHandler.ts
```diff
@@ -0,0 +1,312 @@
...
+
+    let maxLevel = -1;
+    for (const level of levels) {
+        if (totalCount >= level) {
+            maxLevel = level;
+        } else {
+            return maxLevel;
+        }
+    }
+    return -1;
```
jackstenglein, line 200: Change this to `return maxLevel;`

frontend/src/components/profile/info/BadgeCard.tsx
jalpp: fixed

frontend/src/components/profile/info/BadgeCard.tsx
jalpp: You can take a look at my current imp, basically I kept the finding and replacing but added the dep list

frontend/src/components/profile/info/BadgeCard.tsx
jalpp: yea true I removed it for now

frontend/src/components/profile/info/BadgeCard.tsx
jalpp: yea this was from start I fixed it

frontend/src/components/profile/info/BadgeCard.tsx
jalpp: I removed this, I thought about it and tool tip plus this looked weird especially since other badges didn't have it, plus people, can just click on the badge and see it in big

frontend/src/components/profile/info/badgeHandler.ts
jalpp: sure

frontend/src/components/profile/info/badgeHandler.ts
jalpp: yea true

frontend/src/components/profile/info/badgeHandler.ts
jalpp: yup

frontend/src/components/profile/info/badgeHandler.ts
jalpp: yea I changed isRare to isMaxbadge and added the getBadgeGlow method which returns the glow for only max badges, if this rule changes one can at another method and add it inside getBadgeGlow

frontend/src/components/profile/info/badgeHandler.ts
jalpp: yup good catch

frontend/src/components/profile/info/badgeHandler.ts
jalpp: yea I fixed this

frontend/src/components/profile/info/BadgeCard.tsx
jalpp: yea so what I did is I kept the title and the message default to the MUI template because it looks good, The badges glow color rather than using a theme provider I made glow colors so they fit both in dark and light mode. the nonmax badges now are maroon as shown on both. Same for other glow colors

frontend/src/components/profile/info/badgeHandler.ts
jalpp: Yea I was trying to be little to smart there ;)

frontend/src/components/profile/info/badgeHandler.ts
jalpp: yea I added the dash which looks good, was going to add | but it looked weird so dash for now

frontend/src/components/profile/info/badgeHandler.ts
jalpp: yea I changed BADGE -> BadgeType, and also PolgarMateOne, PolgarMateTwo, etc etc

frontend/src/components/profile/info/badgeHandler.ts
jalpp: yea again trying to be a little too smart there ;) I removed it

jackstenglein (approved):

PR #1228, opened 2025-02-15
feat(badges): add badge cabinet, move cohort badges into CustomBadge

jalpp:


frontend/src/components/profile/info/BadgeCard.tsx
```diff
@@ -19,13 +21,28 @@ import Image from 'next/image';
...
 export const BadgeCard = ({ user }: { user: User }) => {
     const [selectedBadge, setSelectedBadge] = useState<Badge | undefined>(undefined);
+    const [isViewAllModalOpen, setIsViewAllModalOpen] = useState(false);
     const badgeData: Badge[] = getEligibleBadges(user);
     const [previousBadgeData, setPreviousBadgeData] = useState<Badge[]>(badgeData);
+    const [badgeCategory, setBadgeCategory] = useState('current');
+    const ineligibleGameBadges: JSX.Element[] = [];
+    const ineligibleAnnoBadges: JSX.Element[] = [];
+    const ineligiblePolgarBadges: JSX.Element[] = [];
+    const inelgibleCohortBadges: JSX.Element[] = [];
```
jackstenglein, line 45: This should be `ineligibleCohortBadges`

frontend/src/components/profile/info/badgeHandler.ts
```diff
@@ -186,6 +196,14 @@ function getBadgeMessage(level: number, badge: BadgeType): string {
 
     return msg;
 }
+/**
+ * gets the cohort badge message
+ * @param cohort the cohort
+ * @returns the message
+ */
+function getCohortBadgeMessage(cohort: string): string {
+    return `Congratulations! You have graduated to ${cohort}`;
```
jackstenglein, line 205: You get the cohort badge when you graduate from a cohort, not when you move into that cohort

frontend/src/components/profile/info/badgeHandler.ts
```diff
@@ -197,6 +215,15 @@ function getBadgeTitle(level: number, badge: BadgeType): string {
     return `${BADGE_TITLE[badge]} - ${level}`;
 }
 
+/**
+ * gets the cohort badge title
+ * @param cohort the cohort
+ * @returns the badge title
+ */
+function getCohortBadgeTitle(cohort: string): string {
+    return `Cohort ${cohort}`;
```
jackstenglein, line 224: This should say `Graduated from ${cohort}` to match what we currently have.

frontend/src/components/profile/info/badgeHandler.ts
```diff
@@ -217,6 +244,15 @@ function getBadgeGlow(level: number, badge: BadgeType): string | undefined {
     return isMaxBadge(level, badge) ? BADGE_MAX_GLOW_COLOR[badge] : undefined;
 }
 
+/**
+ * Gets the cohort badge glow color
+ * @param cohort the cohort
+ * @returns the glow color
+ */
+function getCohortBadgeGlow(cohort: string): string {
```
jackstenglein, line 252: Why do we need this function if all it does is return the color from the map? Why not just use the map directly?

frontend/src/components/profile/info/badgeHandler.ts
```diff
@@ -240,6 +276,22 @@ function getEligibleBadgeInfo(user: User, badge: BadgeType): Badge | undefined {
...
+ * @param cohort the cohort
+ * @returns the badge
+ */
+export function getCohortBadge(cohort: string): Badge {
+    const cohortBadge: Badge = {
+        image: getCohortBadgeImage(cohort),
+        title: getCohortBadgeTitle(cohort),
+        message: getCohortBadgeMessage(cohort),
+        glowHexcode: getCohortBadgeGlow(cohort),
+    };
```
jackstenglein, line 290: Honestly none of these functions seem particularly useful, and we could just inline it all here in this object

frontend/src/components/profile/info/BadgeCard.tsx
```diff
@@ -128,14 +151,70 @@ export const BadgeCard = ({ user }: { user: User }) => {
...
             <Card>
-                <CardHeader title='Badges' />
+                <Stack
+                    direction='row'
+                    justifyContent='space-between'
+                    alignItems='center'
+                    px={2}
+                    pt={2}
+                >
+                    <CardHeader title='Badges' sx={{ p: 0 }} />
```
jackstenglein, line 208: I think we may want more padding between the title and the badges. It looks a bit cramped, especially if the user has graduated, and the glow is almost on top of the title:

Screenshot 2025-02-15 at 10 30 05 AM

frontend/src/components/profile/info/BadgeCard.tsx
```diff
@@ -232,6 +311,127 @@ export const BadgeCard = ({ user }: { user: User }) => {
...
+            >
+                <DialogTitle
+                    sx={{
+                        fontSize: '1.5rem',
+                        fontWeight: 'bold',
+                        textAlign: 'center',
+                        position: 'relative',
+                    }}
+                >
+                    Badge Cabinet
```
jackstenglein, line 337: Let's have this say `All Badges` instead

jackstenglein (changes requested):

frontend/src/components/profile/info/BadgeCard.tsx
jackstenglein: Actually, I don't like this construction where we have an array of ineligible badges for each category. As we add more categories, it will get rather unruly. It also leads to a rather weird UI on the all badges dialog: I see all the badges I've earned together, but when I view a specific category, I only see the unearned ones. I don't see my earned badges in the context of their neighbors.

Additionally, in the badge cabinet, I can't see how to earn a badge without hovering over it, which is annoying if I am trying to look at how to earn all the badges. Also, I know we don't currently show badges on mobile, but we will eventually want to, and mobile doesn't support tooltips very well, so mobile currently would have no way to see how to unlock the badges.

I think we should do something more like the Steam achievement list:

Screenshot 2025-02-15 at 10 39 44 AM

So a scrollable list of the badges, where the badge image is on the left, and the description of how to unlock it is next to it (for now, we don't need the date unlocked or the percentage of people who have it). If the badge is not earned yet, then the image will be gray like you currently have it.

I think we can still separate the badges out by category in the badge cabinet like you currently have it with the dropdown. The dropdown should have the following categories:

- Achieved Badges
- All Badges
- Graduations
- Polgar Mates
- ... etc

Achieved Badges will only have the unlocked badges. Every other category will have unlocked and locked together in a list, with the locked images grayed out. All badges could probably have section headers as well for each category, although that isn't strictly necessary. Let's make sure that the category list in the dropdown follows the order that we display the badges in the badge card.

Finally, this setup will allow you to remove all of these ineligible arrays as well as the `getEligibleBadges` function and combine them into a single `getBadges(user)` function. This function returns an array of all the possible badges. Each badge should have a `category` field set to an enum with options like `Graduation`, `Polgar`, etc. If the user has earned the badge, then the badge has an `isEarned` field set to true. When displaying the badges in the badge card, you loop through all the badges and render only ones where `isEarned` is true. In the badge cabinet, you loop through all the badges and render the badge with an unlocked icon if `isEarned` is true and a locked icon if it is false. If a user selects a category in the dropdown, then you loop through all the badges and render it only if its category matches the selected category.

frontend/src/components/profile/info/BadgeCabinetDialog.tsx
```diff
@@ -0,0 +1,190 @@
...
+import CustomBadge from './CustomBadge';
+
+interface BadgeCabinetDialogProps {
+    isOpen: boolean;
+    onClose: () => void;
+    badgeCategory: string;
+    setBadgeCategory: (category: string) => void;
+    allBadges: Badge[];
+    allGradBadges: Badge[];
+    handleBadgeClick: (badge: Badge) => void;
```
jackstenglein, line 26: I don't think we need `handleBadgeClick` in this dialog. Opening dialogs on top of dialogs is not a good UI, and this dialog will already show the badge image, the description and the requirements to get it, so there is no reason to open the other dialog from this one.

jackstenglein (approved):

Issue #580, opened 2024-06-17
Allow disabling the graduation/demotion dialog


jackstenglein:

jackstenglein:
https://discord.com/channels/951958534113886238/1026952329804644372/1250077129920741492

morganbp:
I can take a look at this task if thats ok for you.
Is it an idea to have a "Don't show again" button on the dialog which disables the alert for a given amount of time or until next graduation?

jackstenglein:
@morganbp Sure, thanks for helping! I think a "Remind Me Later" button which hides it for a month would be good.

jalpp:
hey @morganbp whats your update on this? Long time no see?

morganbp:
Which version of serverless are you running on locally? I do struggle a bit when I try to deploy the serverless-compose.yml

jackstenglein:
I am using serverless 3.38.0. I had a lot of trouble with serverless v4 when it was released, and then gave up on migrating since we want to move away from serverless anyway.

I did recently add some better support for starting up backend development if you run `sls deploy --stage simple`. The simple stage will skip deploying some resources for more extraneous features.

PR #1254, opened 2025-02-22
Allow disabling the graduation/demotion dialog #580

morganbp:
Backend
- Added HideCohortPromptUntil in SiteNotificationSettings, which stores a datetime iso string

Frontend
- Updated SwitchCohortPrompt to not show the snackbar if current date is before HideCohortPromptUntil

### DIFF
--- a/frontend/src/database/user.ts
+++ b/frontend/src/database/user.ts
@@ -213,6 +213,7 @@ export interface SiteNotificationSettings {
     disableNewFollower: boolean;
     disableNewsfeedComment: boolean;
     disableNewsfeedReaction: boolean;
+    hideCohortPromptUntil?: string;
 }
 
 export type MinutesSpentKey =
@@ -851,6 +852,7 @@ export function shouldPromptGraduation(user?: User): boolean {
     return getCurrentRating(user) >= ratingBoundary;
 }
 
+const ONE_MONTH = 1000 * 60 * 60 * 24 * 30;
 const THREE_MONTHS = 1000 * 60 * 60 * 24 * 90;
 
 /**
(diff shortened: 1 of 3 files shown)


morganbp:
demotionSnackbar
promotionSnackbar

frontend/src/database/user.ts
```diff
@@ -894,6 +896,56 @@ export function shouldPromptDemotion(user?: User): boolean {
...
+    if(!hideUntilDate){
+        return false;
+    }
+    
+    const now = new Date().getTime();
+    return now < hideUntilDate;
+}
+
+/**
+ * Creates a partial user object wher hideCohortPrompt is one month (30 days) after today.
```
jackstenglein, line 925: nit: there are a couple typos in this comment. `wher`, `sendt` and `consisten`

frontend/src/database/user.ts
```diff
@@ -894,6 +896,56 @@ export function shouldPromptDemotion(user?: User): boolean {
...
+    siteNotificationSettings ??= {
+        disableGameComment: false,
+        disableNewFollower: false,
+        disableNewsfeedComment: false,
+        disableNewsfeedReaction: false,
+    }
+    const oneMonthForward = new Date();
+    oneMonthForward.setTime(new Date().getTime() + ONE_MONTH);
+    return {
+        notificationSettings:{
```
jackstenglein, line 940: I think the way our backend's user update function works, this may actually wipe their discord and email notification settings and change them back to the default, since you are only including siteNotificationSettings in this object.

jackstenglein (changes requested):

jackstenglein:
@morganbp Thanks for the PR. This looks pretty good, just had a few minor comments. Would you also mind running `npm run format` in the frontend directory after you've addressed the comments? Thanks!

PR #1291, opened 2025-03-01
feat(badges): adding tactics champion badge

jalpp:


frontend/src/components/profile/info/BadgeCard.tsx
```diff
@@ -54,6 +60,11 @@ export const BadgeCard = ({ user }: { user: User }) => {
         }
     }, [earnedBadges, previousEarnedBadges, setSelectedBadge, setPreviousEarnedBadges]);
 
+    if (!isProvisional && currentTacticsRating.overall > minTacticsRating) {
```
jackstenglein, line 63: This should probably be handled inside `getBadges` in `badgeHandler.ts`. I think as you have it here, the tactics champion badge won't show up in the Badge Cabinet.

PR #1419, opened 2025-04-30
fix(discord & rr): fix old discord links, update post migration info,…

jalpp:
… add I read rules checklist in rr register modal


jalpp:
the join the discord part needs to be updated in backend so I guess @jackstenglein you did have to do that

jackstenglein:
@jalpp What are you referring to by "join the discord part"?

PR #1598, opened 2025-07-24
Mobile responsive games tab on profile

Ard2025:


frontend/src/components/profile/directories/AllUploadsSection.tsx
```diff
@@ -34,6 +35,7 @@ export function AllUploadsSection({
     /** Whether to default the navigation menu to open. */
     defaultNavigationMenuOpen?: boolean;
     /** The namespace for the local storage data. */
```
jackstenglein, line 37: Can you move this comment down so that it is above `namespace` again?

frontend/src/components/profile/directories/DirectoriesSection.tsx
```diff
@@ -237,7 +258,8 @@ const DirectorySection = ({
                 )}
 
                 <DataGridPro
-                    autoHeight
```
jackstenglein, line 240: I think we should keep the `autoHeight` parameter for now. I know it is deprecated, but for now I don't want to change that. I'll eventually go through and migrate all the usages of `autoHeight`

frontend/src/components/profile/directories/DirectoriesSection.tsx
```diff
@@ -67,18 +72,33 @@ export const DirectoriesSection = (props: DirectoriesSectionProps) => {
     const directoryId = searchParams.get('directory') || 'home';
     const directoryOwner = searchParams.get('directoryOwner') || props.defaultDirectoryOwner;
 
+    const isMobile = () => {
```
jackstenglein, line 75: Why is this in a function? This breaks react's rules of hooks. It would be better to have this at the top level of the component like:

```typescript
const directoryOwner = searchParams.get('directoryOwner') || props.defaultDirectoryOwner;

const [screenWidth, setScreenWidth] = useState(window.innerWidth);
useEffect(() => {
     const handleResize = () => setScreenWidth(window.innerWidth);
     window.addEventListener("resize", handleResize);
     return () => window.removeEventListener("resize", handleResize);
}, []);
const isMobile = screenWidth <= 1024;

// ... rest of code
```

Or even better, you could use MUI's custom hook useMediaQuery: https://mui.com/material-ui/react-use-media-query/

frontend/src/components/profile/directories/DirectoriesSection.tsx
```diff
@@ -262,18 +284,18 @@ const DirectorySection = ({
...
                     rowSelectionModel={rowSelectionModel}
                     isRowSelectable={isDirectorySelectable}
-                    rowReordering={isAdmin}
+                    rowReordering={isAdmin && !isMobile}
                     onRowOrderChange={handleRowOrderChange}
                     pagination
                     pageSizeOptions={pageSizeOptions}
-                    sx={{ width: 1, ...sx }}
-                    showToolbar
+                    sx={{ width: 1, ...sx, display: 'grid' }}
```
jackstenglein, line 297: Why is `display: 'grid'` necessary? Also, can we move it before the `...sx` prop spread, so that it is possible to override it via `sx` if necessary?

frontend/src/components/profile/directories/DirectoriesSection.tsx
```diff
@@ -288,6 +310,45 @@ const DirectorySection = ({
     );
 };
 
+function ListViewCell(params: GridRenderCellParams) {
+    console.log(params)
```
jackstenglein, line 314: Can this console log be removed?

frontend/src/components/profile/directories/DirectoriesSection.tsx
```diff
@@ -288,6 +310,45 @@ const DirectorySection = ({
...
+    if(params.row.type == "DIRECTORY"){
```
jackstenglein, line 315: This condition should use `===`

jackstenglein (changes requested):
Can you also run `npm run format` in the frontend directory after you have made the changes?

frontend/src/components/profile/directories/DirectoriesSection.tsx
Ard2025: yes

frontend/src/components/profile/directories/DirectoriesSection.tsx
Ard2025: There was some ui bug causing the main content to get under the sidebar on the right. This fixes that.

frontend/src/components/profile/directories/DirectoriesSection.tsx
Ard2025: That is a really good point of feedback 😀. i just refactored it with the usemediaquery to make the code more clean

jackstenglein (approved):

PR #1737, opened 2025-09-01
Prettier cross os compatible

Mudenz:


.prettierrc
```diff
@@ -1,15 +1,16 @@
...
-        }
-    ],
-    "plugins": ["prettier-plugin-organize-imports"]
-}
+{
+    "jsxSingleQuote": true,
+    "printWidth": 100,
+    "singleQuote": true,
+    "tabWidth": 4,
+    "endOfLine": "lf",
```
Mudenz, line 6: only line 6 is changed.

PR #1960, opened 2026-02-16
feat: Shift + D opens hot key setting modal

### LINKED ISSUE #1846: Game Editor - allow user to display pop-up graphic of available hotkeys
Chesscom's analysis board has a nice feature where you can hit shift+D to bring up a list of available hotkeys -

Image

I also like the defaults that they use for 1-7 and then shift 1-7

Currently users can set their own hotkeys so the popup update to respect their choices

jalpp:
fixes #1846


jalpp:
> > @jalpp The dialog only opens if the user is already on the settings tab. It doesn't open from any other tabs. Also, we shouldn't hard code shift+D as the hotkey for opening the dialog, as users may have already set shift+D to other shortcuts. We should add a new shortcut for opening the dialog and set the default to shift+D
> 
> I have fixed it according to this description. I have left modifer has empty for now. Not sure if shift there is needed by default @jackstenglein



> > @jalpp The dialog only opens if the user is already on the settings tab. It doesn't open from any other tabs. Also, we shouldn't hard code shift+D as the hotkey for opening the dialog, as users may have already set shift+D to other shortcuts. We should add a new shortcut for opening the dialog and set the default to shift+D
> 
> I have fixed it according to this description. I have left modifer has empty for now. Not sure if shift there is needed by default @jackstenglein

I added regardless

PR #1986, opened 2026-02-20
feat: chess cloud base eval 

jalpp:


frontend/src/board/pgn/explorer/ChessDb.tsx
```diff
@@ -0,0 +1,105 @@
...
+    fen: string;
+}
+
+export function ChessDBTab({ moves, loading, error, requestAnalysis }: ChessDBTabProps) {
+    const { chess } = useChess();
+    const reconcile = useReconcile();
+
+    if (loading) return <CircularProgress sx={{ mt: 2 }} />;
+
+    if (error)
```
jackstenglein: Please add braces to any if statement that isn't a single line

jackstenglein (changes requested):

frontend/src/board/pgn/pgnText/engine/EvaluationSection.tsx
```diff
@@ -93,3 +113,177 @@ export const EvaluationSection = ({
...
+    fontSize: '0.9rem',
+    cursor: 'pointer',
+    '&:hover': {
+        color: theme.palette.primary.main,
+    },
+}));
+
+function cloudPvToMoves(fen: string, pvUci: string[]): (Move | null)[] {
+    const game = new Chess({ fen });
+    return pvUci.map((uci) => {
```
jackstenglein: Why are we doing this and not just using the pvSAN field returned by cloud db?

frontend/src/board/pgn/pgnText/engine/EvaluationSection.tsx
```diff
@@ -93,3 +113,177 @@ export const EvaluationSection = ({
...
+        if (move.color === Color.black) {
+            label += '..';
+        }
+        label += ' ';
+    }
+    label += move.san;
+    return label;
+}
+
+function CloudEvalSection({ pv, loading }: { pv: ChessDbPv | null; loading: boolean }) {
```
jackstenglein: This component has a bunch of copied logic and functions from the LineEval.tsx file. It would be better to instead convert the cloud DB result into a LineEval and pass it to the LineEvaluation component and reuse all that existing logic. You can add an icon parameter to the LineEvaluation component in order to display the cloud icon.

jackstenglein (changes requested):

PR #1989, opened 2026-02-21
Fix #1862: Add delete confirmation and 404 handling for deleted games

### LINKED ISSUE #1862: Easy to accidentally delete games from multiple folders
**📓 Steps to reproduce:** 
  > I created a new game without sticking it in a folder first.  Then I saved in into the specific folder I wanted and deleted the original save that was not in a folder to clean things up. It seems like the one in the folder was maybe just a link to the original, rather than a separate save, as it now says 500 Internal Server Error when I try to open from the folder.  I can't find a spot to recover the deleted version and, if working as intended, don't feel like it was clear that the second save was dependent on the first remaining intact.
🔨 Platform: chessdojo.club
📱 Device Type: Desktop
🌐 Browser Type: Chrome

👤 Author name: Doozre
[Discord URL](https://discord.com/channels/951958534113886238/1299913797045325957/1438327015009681634)

Salocin11:
Changes:
- Added typed "delete" confirmation to DeleteGamesDialog to prevent accidental deletions.
- Updated GamePage to redirect to 404

Result:
- Deleting a game now requires typing "delete".
- Opening a deleted game URL shows a Not Found page instead of a 500 error.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 node_modules/
 .todo
 /.vs
+.cursor/
--- a/backend/database/notification.go
+++ b/backend/database/notification.go
@@ -395,6 +395,66 @@ func (repo *dynamoRepository) ListNotifications(username string, startKey string
 	return notifications, lastKey, nil
 }
 
+// DeleteAllNotifications removes all notifications for the specified username from the database.
+func (repo *dynamoRepository) DeleteAllNotifications(username string) error {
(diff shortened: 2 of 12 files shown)


jackstenglein:
@Salocin11 Thanks for opening this PR. However I merged #1990 instead of this one because that one had a test as well. For future issues, please comment on it if you are working on it, that way we don't duplicate work. Thanks again!

Issue #1903, opened 2026-01-03
Cannot graduate if you dismiss graduation task Tech_BUG#443


jackstenglein:
**📓 Steps to reproduce:** 
  > Go to training page
dismiss graduation task from top pain
look for graduation task in complete list below
realize you are an idiot for dismissing graduation task
🔨 Platform: https://www.chessdojo.club/profile?view=progress
📱 Device Type: Desktop
🌐 Browser Type: Edge

👤 Author name: Eric The Nerd
[Discord URL](https://discord.com/channels/951958534113886238/1299913797045325957/1457098638810415309)

jackstenglein:
To fix this, we should probably add a fake graduation task to the full training plan, which should be disabled if you haven't reached the required rating yet.

PR #1992, opened 2026-02-21
Changed Full Training Plan to have a locked graduation task when the …

mindvoxel:
…user is not eligible. Behaves same as "daily" graduation task when user is eligible for graduation. (Opens dialogue box, upsell logic, etc.).

adresses #1903

@jackstenglein let me know if this is what you had in mind for the "fake task" 

https://github.com/user-attachments/assets/09d96e45-622f-4db8-8f54-d2a9cd7ad18b


frontend/src/components/profile/trainingPlan/full/FullTrainingPlanSection.tsx
```diff
@@ -204,20 +201,6 @@ export function FullTrainingPlanSection({
                         Add Custom Task
                     </Button>
                 )}
-
-                {isFreeTier &&
```
jackstenglein: This also shouldn't be removed

frontend/src/components/profile/trainingPlan/full/FullTrainingPlanGraduationItem.tsx
```diff
@@ -0,0 +1,149 @@
...
+    user,
+    cohort,
+    isCurrentUser,
+}: FullTrainingPlanGraduationItemProps) {
+    const isFreeTier = useFreeTier();
+    const [dialogOpen, setDialogOpen] = useState(false);
+    const [upsellOpen, setUpsellOpen] = useState(false);
+
+    const canGraduate = shouldPromptGraduation(user);
+    const disabled = !canGraduate;
```
jackstenglein: Can we make sure this is always enabled if the user's preferred rating system is a custom system? Custom users currently can't graduate and it would be nice to fix that as part of this change.

PR #2006, opened 2026-02-23
feat: add task timer buttons and fix global state overlap

H4ROOON:
added the start/stop timer buttons directly inside the task modal. i wired them to existing global useTimer from the navbar so the state stays perfectly synced. clicking "Stop Timer" automatically opens the update progress view with the tracked time pre-filled.

i also fixed a small UX edge case I noticed while testing: if you tracked time on task A, submitted your progress, and then started the timer for task B, the global clock wouldnt reset and would double-count your time.


frontend/src/components/profile/trainingPlan/TaskDialog.tsx
```diff
@@ -296,8 +300,32 @@ function DetailsDialog({ task, onClose, cohort, setView }: DetailsDialogProps) {
...
+                    <Button 
+                        variant="contained" 
+                        color="error" 
+                        startIcon={<Stop />} 
+                        onClick={() => {
+                            onPause();
+                            setView(TaskDialogView.Progress);
+                        }}
+                    >
+                        Stop Timer ({formatTime(timerSeconds)})
```
jackstenglein: nit: I think this should say pause, as stop might imply that you will completely reset the time

jackstenglein:
@H4ROOON Would it be possible to render a new component when the timer is stopped that wraps the dialog in a timeline provider? Then the timeline will only be loaded when the user stops the timer?

Issue #1996, opened 2026-02-22
feat: Time Management Rating


jessekraai:
This is a major task that we have not accomplished despite in being in the docket for a long time. It shouldn't be that hard. Once a player has at least 10 games in their MyGames folder with a time management rating then we average those and present this number below their preferred rating beneath their heatmap. After 10 each new time management rating is scored as a draw against their present rating, using the uscf rating system. Ie. you have an average rating of 1450 after 10 games and your new game has one of 1800. That counts as a draw against an 1800.

Dblike:
@jessekraai I used this issue as an opportunity to familiarize myself with the backend, and have taken a crack at implementing it though there a couple outstanding questions on my PR

jessekraai:
I still couldn't find the questions on my own. Not even AI could help me.

To answer 1) simply show the rating. should be labeled Time Management Rating and have a clock icon.
2) Jack will have to answer this one.

jessekraai:
Let's move on this! @jackstenglein

PR #2007, opened 2026-02-23
feat(#1996): add time management rating for game uploads

Dblike:
**ticket:** https://github.com/jackstenglein/chess-dojo/issues/1996

### overview

Adds a per-game "time management rating" (0–3000) computed from clock annotations in the PGN, persisted on game records, and aggregated into a user-level rating on the profile. Only classical games (30+ min) with real `{[%clk ...]}` data and 5+ moves are scored. Includes a backfill script for existing games.

### before + after

| before | after |
| --- | --- |
| Profile shows no time management rating | "Time Management" row in Dojo Score Card with rating + provisional indicator |
| Game records have no TM fields | Games store `timeManagementRatingWhite` / `Black` |
| No user-level TM aggregate | Dedicated `User.timeManagementRating` field (not in RatingSystem enum) |

### key changes

- Moved clock rating calculation from frontend to `common/src/ratings/clockRating.ts` (single source of truth)
- Per-game TM ratings calculated and persisted on game create/update
- Incremental user aggregation: running average (provisional, < 10 games) → USCF Elo draw adjustment (K=32)
- TM rating stored as a dedicated top-level `User.timeManagementRating` field — not in the `RatingSystem` enum, since it's a computed internal metric, not a user-entered external rating
- Profile display with provisional indicator in DojoScoreCard
- Backfill script (`backfillTimeManagement.ts`) — idempotent, single-pass scan

<details>
<summary><b>Design decisions</b></summary>

**Dedicated field, not a RatingSystem member** — `RatingSystem` represents external, user-entered rating systems (Chess.com, Lichess, FIDE, etc.). TimeManagement is a computed aggregate with no username, no graduation boundary, and shouldn't appear in rating selectors or search fields. A top-level `User.timeManagementRating` field avoids needing filters everywhere the codebase iterates `Object.values(RatingSystem)`.

**Store both white and black per-game ratings** — cheap to compute in one pass, avoids ambiguity about which side the owner played, enables future comparison features.

**Incremental aggregation (no history array)** — DynamoDB User records are under size pressure (~400KB limit). A `gameRatings[]` array has unbounded growth and race conditions under concurrent Lambdas. Instead we store `currentRating` + `numGames` and update incrementally. Per-game ratings on Game records are the durable source of truth — the backfill script doubles as repair.

**Skip games without clock annotations** — without real `clk` data, the fallback produces a flat line and a meaningless 0 rating. The `hasClockData` guard prevents pollution.

**No aggregate update on game edit** — PGN edits are rare, correction is complex, drift is at most one game. Backfill script corrects accumulated drift.

</details>

<details>
<summary><b>Open questions</b></summary>

**Q1: How should time management be displayed on the profile?**

Currently rendered as a simple icon + text + provisional chip in DojoScoreCard. Every other row in the card uses `ScoreboardProgress` with a progress bar. Options: keep as-is, add progress bar (e.g. `3/10` toward established), use rating boundaries like the main rating system, or extract to its own component. Happy to implement whichever fits.

Time Management Rating

**Q2: Is incremental aggregation acceptable, or should we store a full history array?**

Currently we store only `currentRating` + `numGames` on the User record and update incrementally. This avoids unbounded array growth and race conditions under concurrent Lambdas, but means the aggregate can't be recalculated from the User record alone — per-game ratings on Game records are the source of truth, and the backfill script serves as the repair path. If a full history array is preferred (e.g. for charting rating over time), we'd need a cap or pagination strategy to stay within DynamoDB's 400KB item limit.

</details>


Dblike:
> @Dblike Thanks for the PR! I haven't reviewed the code yet (planning to get to it tomorrow) but this line stood out to me in the description:
> 
> > No aggregate update on game edit — PGN edits are rare, correction is complex, drift is at most one game. Backfill script corrects accumulated drift.
> 
> Does this mean if I import my game, then edit it later, my overall time management rating isn't updated? I think that presents a big problem if so. Imagine I start entering my OTB game, and only enter clock data for half the game. Then a day later, I come back and enter the rest of the clock data. My time management rating for that game will be drastically different, so my overall time management rating should also change.

You're right — the original PR didn't update the aggregate on game edit. This is now fixed. When a game's PGN is edited (or orientation changes), we rebuild the user's aggregate TM rating from scratch by querying all their games via OwnerIdx and reprocessing. This handles exactly the scenario you described — entering clock data across multiple sessions. The rebuild is cheap (1 GSI query + 1-2 BatchGetItems + 1 UpdateItem) and only triggers when TM-relevant fields actually change, which is rare.

> Your open questions:
> 
> > Q1: How should time management be displayed on the profile?
> 
> Similar to what you have, but let's have the rating right-aligned so there is space between it and the label. Also if provisional, let's add a `?` after the rating instead of the chip.

Rating is now right-aligned with a ? suffix for provisional (< 10 games) instead of the chip.

> 
> > Q2: Is incremental aggregation acceptable, or should we store a full history array?
> 
> I don't think we should store a full history array. The team had previously discussed a moving average or exponential moving average of the most recent X games, where the value of X had not yet been figured out. What does incremental aggregation mean in this case? Is it just a simple average?

The aggregation is not a simple average — it's a two-phase approach based on Jesse's comments in #1996:

Provisional (< 10 games): Running average. Each game's TM score is weighted equally. This gives a stable baseline quickly.
Established (≥ 10 games): USCF-style Elo draw adjustment (K=32). Each new game's TM score is treated as a "draw opponent" — if it matches the current rating, no change; higher pulls it up, lower pulls it down. The Elo curve bounds the impact of outliers.
Compared to a moving average of the last X games, this approach avoids storing individual game scores on the user record — we only store { currentRating, numGames }. The tradeoff is it's not reversible (can't "undo" one game's contribution), which is why game edits trigger a full rebuild.

If you'd prefer a moving average or EMA instead, happy to discuss what X should be and switch approaches. The per-game TM scores are already stored on each game record either way, so the foundation supports both.

backend/pgnService/game/create.ts
```diff
@@ -482,3 +500,167 @@ export async function createTimelineEntry(game: Game) {
...
+/**
+ * Rebuilds a user's aggregate time management rating from all their games.
+ * Used when a game is edited and its TM rating changes, since the incremental
+ * aggregate model is not reversible.
+ * @param owner The username of the game owner.
+ */
+export async function rebuildUserTimeManagementRating(owner: string): Promise<void> {
+    try {
+        // Query OwnerIdx for all game keys belonging to this owner
+        let startKey: Record<string, import('@aws-sdk/client-dynamodb').AttributeValue> | undefined;
```
jackstenglein: This import should be at the top of the file

backend/pgnService/game/create.ts
```diff
@@ -482,3 +500,167 @@ export async function createTimelineEntry(game: Game) {
...
+        const gameKeys: { cohort: string; id: string }[] = [];
+
+        do {
+            const result = await dynamo.send(
+                new QueryCommand({
+                    TableName: gamesTable,
```
jackstenglein: Jesse will definitely want to limit this search to only games in the user's mygames folder. I'm also worried that this may introduce a lot of latency for users with many games. We may need to trigger a separate lambda to do this update in the background. How many games did you test this on?

backend/pgnService/game/create.ts
```diff
@@ -482,3 +500,167 @@ export async function createTimelineEntry(game: Game) {
...
+    }
+}
+
+/**
+ * Rebuilds a user's aggregate time management rating from all their games.
+ * Used when a game is edited and its TM rating changes, since the incremental
+ * aggregate model is not reversible.
+ * @param owner The username of the game owner.
+ */
+export async function rebuildUserTimeManagementRating(owner: string): Promise<void> {
```
jackstenglein: This function isn't used in this file and doesn't make a lot of sense to include in the create API file. We should probably move it to one of the timeManagement related files

backend/pgnService/game/timeManagement.test.ts
```diff
@@ -0,0 +1,102 @@
...
+
+const noTimeControlPgn = `[Event "Test"]
+
+1.e4 {[%clk 1:29:40]} e5 {[%clk 1:29:30]} 2.Nf3 {[%clk 1:29:10]} Nc6 {[%clk 1:28:50]} 1-0`;
+
+test('returns ratings for classical game with clocks', () => {
+    const chess = new Chess({ pgn: classicalPgnWithClocks });
+    const ratings = calculateTimeManagementRatings(chess);
+
+    assert.isDefined(ratings.white, 'white rating should be defined');
```
jackstenglein: Can we get some games from prod where we already know the time management rating and verify that they exactly match in these tests, rather than just checking that they are in the range [1, 3000]?

backend/pgnService/game/timeManagement.ts
```diff
@@ -0,0 +1,59 @@
...
+    const initialSeconds = timeControls[0].seconds ?? 0;
+    const moves = chess.history();
+
+    const whiteClock: ClockDatum[] = [{ seconds: initialSeconds }];
+    const blackClock: ClockDatum[] = [{ seconds: initialSeconds }];
+    let hasClockData = false;
+
+    for (let i = 0; i < moves.length; i += 2) {
+        const whiteSeconds = clockToSeconds(moves[i]?.commentDiag?.clk);
+        if (whiteSeconds !== undefined) hasClockData = true;
```
jackstenglein: Use braces for if statements unless there is a single `return`, `break` or `continue`

backend/pgnService/game/timeManagementAggregation.test.ts
```diff
@@ -0,0 +1,136 @@
...
+    calculateDrawEloAdjustment,
+    MIN_GAMES_FOR_ELO,
+    updateTimeManagementAggregate,
+} from '@jackstenglein/chess-dojo-common/src/ratings/timeManagement';
+import { calculateTimeRating, ClockDatum } from '@jackstenglein/chess-dojo-common/src/ratings/clockRating';
+import { assert, test } from 'vitest';
+
+// --- clockRating.ts tests (Layer 1: pure calculation) ---
+
+test('calculateTimeRating returns undefined for too few moves', () => {
```
jackstenglein: The tests that are verifying behavior from `common` should be in the `common` directory

backend/pgnService/game/timeManagementAggregation.test.ts
```diff
@@ -0,0 +1,136 @@
...
+});
+
+test('updateTimeManagementAggregate is provisional for < 10 games', () => {
+    let agg = updateTimeManagementAggregate(undefined, 2000);
+    for (let i = 1; i < MIN_GAMES_FOR_ELO - 1; i++) {
+        agg = updateTimeManagementAggregate(agg, 2000);
+    }
+
+    assert.equal(agg.numGames, MIN_GAMES_FOR_ELO - 1);
+    assert.isBelow(agg.numGames, MIN_GAMES_FOR_ELO);
```
jackstenglein: This second assert must be true if the previous was true. Let's delete this one

backend/pgnService/game/timeManagementAggregation.test.ts
```diff
@@ -0,0 +1,136 @@
...
+    for (let i = 1; i < MIN_GAMES_FOR_ELO; i++) {
+        agg = updateTimeManagementAggregate(agg, 2000);
+    }
+    assert.equal(agg.numGames, MIN_GAMES_FOR_ELO);
+    assert.equal(agg.currentRating, 2000);
+
+    // Game 11 at 2500 should use Elo draw adjustment, not average
+    const afterElo = updateTimeManagementAggregate(agg, 2500);
+    assert.equal(afterElo.numGames, MIN_GAMES_FOR_ELO + 1);
+    assert.isAbove(afterElo.currentRating, 2000, 'Elo should increase toward 2500');
```
jackstenglein: Can we calculate what it should be and check for that exact number here?

backend/pgnService/game/timeManagement.ts
```diff
@@ -0,0 +1,59 @@
...
+    black?: number;
+}
+
+/**
+ * Extracts clock data from a parsed Chess instance and calculates
+ * time management ratings for both sides.
+ * @param chess The parsed Chess instance with move history.
+ * @returns An object with optional white and black time management ratings.
+ */
+export function calculateTimeManagementRatings(chess: Chess): TimeManagementRatings {
```
jackstenglein: It feels like this function should be in common/ratings/timeManagement.ts or common/ratings/clockRating.ts

common/src/ratings/timeManagement.ts
```diff
@@ -0,0 +1,75 @@
...
+}
+
+/**
+ * Calculates the updated time management rating after a new game, treating it
+ * as a draw against the new game's time management rating using the USCF Elo system.
+ * @param currentRating The player's current time management rating.
+ * @param gameRating The time management rating from the new game.
+ * @returns The updated rating, rounded to the nearest integer.
+ */
+export function calculateDrawEloAdjustment(
```
jackstenglein: I would call this function something like `updateTimeManagementRating`. The current name isn't clear that it's related to the time management, and callers of the function should not need to know that it uses a USCF Elo draw system to update the rating. Adjustment also makes it sound like it returns the difference between the old rating and new rating.

jackstenglein (changes requested):

backend/pgnService/game/create.ts
Dblike: I limited the query to just a user's mygames folder. I tested it on up to 500 games and latency was only ~150ms.

jackstenglein (approved):

Issue #2014, opened 2026-02-24
feat: optimize local Stockfish engine's speed


jalpp:
Currently, the local engine computes and forgets the eval completely, we can actually store it in browser cache and optimize the speed to O(1) look up. This is super useful when a user checks the same variation of a game for a lot of games. For that variation lookup, it will be so fast that the user will think it's engine-calculated that fast. I will get on this

PR #2028, opened 2026-02-26
feat: make local stockfish blazing fast 

jalpp:
makes it faster than bolt #2014


frontend/src/stockfish/hooks/useEval.ts
```diff
@@ -72,14 +92,22 @@ export function useEval(enabled: boolean, engineName?: EngineName): PositionEval
...
-                    ...savedEvals.current,
-                    [fen]: { ...rawPositionEval, engine: engineName },
-                };
-            } catch (err) {
-                if (err !== E_CANCELED) {
-                    throw err;
+                // Guard: ensure the position hasn't changed while engine was running.
+                // If the user moved away and back, fen may differ — don't cache a
+                // result that belongs to a position we're no longer on.
+                if (rawPositionEval.lines[0]?.fen !== fen) {
```
jackstenglein: Is this condition ever actually true in your testing? I would expect this to be an issue in the current implementation if so.

frontend/src/stockfish/hooks/evalCache.ts
```diff
@@ -0,0 +1,37 @@
+import { openDB } from 'idb';
+import { EngineName, SavedEval } from '../engine/engine';
+
+const DB_NAME = 'chessEngineEvals';
+const STORE_NAME = 'evals';
+const DB_VERSION = 1;
+
+async function getDb() {
```
jackstenglein: When does this cache ever get cleared? How big can it grow? What happens if it runs out of space?

jackstenglein (changes requested):

frontend/src/api/cache/chessdb.ts
```diff
@@ -184,12 +76,7 @@ async function evictIfNeeded(db: IDBPDatabase, incomingBytes: number): Promise<v
  * @returns The cached {@link ChessDbCacheEntry}, or `undefined` on a cache miss.
  */
 export async function getChessDbCache(fen: string): Promise<ChessDbCacheEntry | undefined> {
-    const db = await getDb();
-    const value = (await db.get(STORE_NAME, fen)) as ChessDbCacheEntry | undefined;
-    if (value) {
-        void touchMeta(db, fen, measureBytes(value));
-    }
-    return value;
+    return chessDbCache.get(fen);
```
jackstenglein, line 79: I should have commented this on your last PR, but should we be normalizing this FEN? If someone gets to the same board position but with different move number, then an unnormalized FEN wouldn't get the caching benefits. We should probably also normalize the FENs with the SF cache

frontend/src/api/cache/chessdb.ts
```diff
@@ -208,24 +95,24 @@ export async function getChessDbCache(fen: string): Promise<ChessDbCacheEntry |
  * @param entry - The {@link ChessDbCacheEntry} to cache.
  */
 export async function setChessDbCacheEntry(fen: string, entry: ChessDbCacheEntry): Promise<void> {
-    const db = await getDb();
+    const db = await chessDbCache.getDb();
```
jackstenglein, line 98: Why is this not written using the new LruCache API?

Can't this entire function just be:

```typescript
const existing = await chessDbCache.get(fen);
const merged = {...existing, ...entry};
await chessDbCache.set(fen, merged);
```

frontend/src/stockfish/hooks/useEval.ts
```diff
@@ -36,25 +39,35 @@ export function useEval(enabled: boolean, engineName?: EngineName): PositionEval
...
-                setCurrentPosition(savedEval);
+                setCurrentPosition(memHit);
+                return;
+            }
+
+            // L2 – IndexedDB (survives page reload; only contains fully-completed evals)
+            const idbHit = await getEvalCache(cacheKey);
+            if (idbHit) {
+                memCache.current[fen] = idbHit;
+                setCurrentPosition(idbHit);
```
jackstenglein: What if this cached version isn't high enough depth compared to what the user is currently trying to analyze? We would never show them anything deeper than what is cached, even if local stockfish passes it

jackstenglein (changes requested):

jackstenglein:
There are also some merge conflicts which need to be resolved

Issue #1532, opened 2025-06-07
Setting min goal for custom tasks TECHGEN#119


jackstenglein:
**💬 Content:** 
  > Hi, it would be useful to be able to set the 'min' goal for custom tasks. 

This is a feature with default tasks like "read silman's endgame part 5" has a goal slider that runs from page 158 to page 212. In custom tasks there doesn't seem to be an option for setting the min. Thanks

👤 Author name: elliot
[Discord URL](https://discord.com/channels/951958534113886238/1026952329804644372/1380801197203652682)

romyers:
I see this issue is rather old, but it doesn't look like this has been done yet. I'd like to start with this.

PR #2036, opened 2026-02-27
feat(#1532): Setting min goal for custom tasks

romyers:
## Summary

feat(#1532). Added an input field for specifying a minimum goal when creating a custom task.

### Frontend:

Added an input field to the CustomTaskEditor labeled 'Starting Point' in which a user can input a value for a task to start from.

Added validators to CustomTaskEditor ensure the starting point is a valid integer on the range [0, Goal).

Added validation to CustomTaskEditor to ensure that 'goal' is a valid integer. Previously, a noninteger goal would just have its fractional part silently discarded.

Added playwright tests and element testIds relevant to the new feature.

### Backend:

Added 'StartCount' attribute to CustomTask in backend/database/requirement.go, configured as an optional pointer to ensure backward compatibility with preexisting database items.

## Type of change

*   [ ] Documentation update

## Testing

## Demo
NewFeature

### DIFF
--- a/frontend/src/components/profile/trainingPlan/CustomTaskEditor.tsx
+++ b/frontend/src/components/profile/trainingPlan/CustomTaskEditor.tsx
@@ -95,12 +95,12 @@ const CustomTaskEditor: React.FC<CustomTaskEditorProps> = ({
         if (cohorts.length === 0) {
             newErrors.cohorts = 'At least one cohort is required';
         }
-        const startCountInt = parseInt(startCount || '0');
-        if (isNaN(startCountInt) || startCountInt < 0) {
+        const startCountInt = Number(startCount || '0');
+        if (!Number.isInteger(startCountInt) || startCountInt < 0) {
             newErrors.startCount = 'Must be a positive integer or empty';
         }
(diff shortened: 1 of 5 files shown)


frontend/src/components/profile/trainingPlan/CustomTaskEditor.tsx
```diff
@@ -92,10 +95,17 @@ const CustomTaskEditor: React.FC<CustomTaskEditorProps> = ({
         if (cohorts.length === 0) {
             newErrors.cohorts = 'At least one cohort is required';
         }
-        const countInt = parseInt(count || '0');
-        if (isNaN(countInt) || countInt < 0) {
+        const startCountInt = Number(startCount || '0');
+        if (!Number.isInteger(startCountInt) || startCountInt < 0) {
+            newErrors.startCount = 'Must be a positive integer or empty';
+        }
+        const countInt = Number(count || '0');
```
romyers: I've changed this to avoid silently truncating noninteger values for 'count,' allowing for the validator to notice them and display an error for the user.

jackstenglein (approved):

PR #2041, opened 2026-03-01
feat: Stockfish code clean up and 18 support

jalpp:
this PR cleans up code to only use Stockfish js port for all fishes, note please test #2040 first

### DIFF
--- a/frontend/src/stockfish/hooks/useEngine.ts
+++ b/frontend/src/stockfish/hooks/useEngine.ts
@@ -12,7 +12,7 @@ export const useEngine = (enabled: boolean, engineName: EngineName | undefined) 
         pickEngine(engineName)
             .then((newEngine) => {
                 setEngine((prev) => {
-                    void prev?.shutdown();
+                    prev?.shutdown();
                     void prev?.stopAllCurrentJobs();
                     return newEngine;
                 });


frontend/src/stockfish/engine/helper.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const isMobileDevice = () =>
+    isIosDevice() || /Android|Opera Mini/i.test(navigator.userAgent);
+
+/**
+ * Passes the given message and params to console.debug if this._debug is true.
+ * @param message The message to pass to console.debug.
+ * @param optionalParams The optionalParams to pass to console.debug.
+ */
+export function debug(message?: unknown, ...optionalParams: unknown[]) {
```
jackstenglein: This function no longer makes sense here. It used to check if `this._debug` was true before logging, but now it isn't part of a class, and doesn't check anything. We should delete this function and just use logger.debug instead.

frontend/src/stockfish/engine/StockfishEngineFactory.ts
```diff
@@ -0,0 +1,32 @@
...
+ * @returns UCIEngineFactory
+ */
+export async function createStockfishEngine(
+    stockfishVersion: EngineName,
+): Promise<UciEngineFactory> {
+    debug(`[Stockfish] Creating engine for version: ${stockfishVersion}`);
+
+    if (!isStockfishSupported()) {
+        const error = `${stockfishVersion} is not supported`;
+        debug(`[Stockfish] ${error}`);
```
jackstenglein: Debug messages should be for detailed messages that help debugging. Errors should use logger.error so they have higher visibility and show up even in prod. It also probably makes more sense to not log anything here and instead log this where the error is caught

frontend/src/stockfish/engine/UciEngineFactory.ts
```diff
@@ -0,0 +1,496 @@
...
+     *
+     * @example
+     * ```ts
+     * const engine = await UciEngineFactory.create(EngineName.Stockfish, '/stockfish.wasm');
+     * ```
+     */
+    public static async create(
+        engineName: EngineName,
+        enginePath: string,
+        customEngineInit?: UciEngineFactory['customEngineInit'],
```
jackstenglein: This parameter is unused outside of this file. Is it still necessary?

PR #2044, opened 2026-03-02
feat: add newsfeed icon button to navbar for quick access

brianestraduh:
Summary
Adds a persistent Newsfeed icon button (Feed icon) to the right side of the navbar, alongside the existing Notification and Help icons. This ensures the newsfeed is always one click away regardless of screen size, rather than being buried in a submenu. On narrower viewports (<595px), the icon gracefully falls back to a menu item in the overflow menu.
Related Issues
Fixes TECHGEN#932
Type of change
[ ] Bug fix (non-breaking change which fixes an issue)
[x] New feature (non-breaking change which adds functionality)
[ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
[ ] Documentation update
Testing
[x] New unit tests (vitest) were added
[ ] New playwright tests were added
[ ] It was not necessary to add tests due to {{reason}}
Demo
The Feed icon appears in the navbar's right-side icon group (next to Notifications and Help). Clicking it navigates to /newsfeed.

### DIFF
--- a/frontend/playwright/tests/e2e/profile/directories/DirectoriesTab.spec.ts
+++ b/frontend/playwright/tests/e2e/profile/directories/DirectoriesTab.spec.ts
@@ -63,7 +63,7 @@ test.describe('Directories', () => {
     });
 
     test('links to game import page', async ({ page }) => {
-        await page.getByRole('button', { name: 'Add' }).click();
+        await page.getByRole('button', { name: 'Add', exact: true }).click();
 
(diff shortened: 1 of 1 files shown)


jackstenglein:
@brianestraduh I don't think we should put this back into the navbar. It's not really important enough to get its own icon on the right. I think we should instead have a small mini newsfeed on the profile which links to the main newsfeed

brianestraduh:
All done here is a screenshot of it on the profile page 
image

Issue #2046, opened 2026-03-02
fix: Handling for a task's 'min goal' in several fields


romyers:
- The ProgressUpdater usually autofills the progress count with the user's last known progress. However, it always autofills zero if there is no progress associated with that task in the user's history, even if the task is supposed to start at a nonzero count (e.g. Polgar M3s start at 3718 but the first progress update autofills 0):
![Image](https://github.com/user-attachments/assets/5408b56c-7a9c-4ede-9f13-b751c890658f)
- The 'min goal' is counted in the progress history for the first progress update, so e.g. if a user starts Polgar M3s and does 10 puzzles, the progress history will show a count of 3728, and the historical total count will also show 3728. If the user does 5 more puzzles, that entry will show a count of 5. This is inconsistent.
![Image](https://github.com/user-attachments/assets/caf385d4-d911-458a-bfc9-ef8695aa4bb4)
- Daily training plan cards always show a 'min goal' of zero until the first progress update for the associated task. E.g. Polgar M3s show '0 / 4462 exercises completed' until the user submits progress. If the user then completes one exercise, it will show '3719 / 4462 exercises completed', which sounds like the user has done 3719 exercises instead of just one.
![Image](https://github.com/user-attachments/assets/7f3f4dd4-0846-4ce7-b9a5-d9dedb0a5ddd)
It would be better to either:
  - offset the goal by startCount to reflect the number of exercises actually done, so that e.g. Polgar M3s will show '0 / 744 exercises completed', or
  - use startCount for the first number when there are no progress updates, and change the explanatory text to match the semantics (e.g. for Polgar M3s, 'Current progress: 3718 / 4462' or simply '3718 / 4462'.

PR #2050, opened 2026-03-03
fix(#2046): Handling for a task's min goal in several fields

romyers:
## Summary

Fix(#2046): 
- The progress update dialog now autofills a task's 'min goal' when no progress updates have been made yet.
change3
- Daily training plan cards now display the count a user has actually done for a task and the total count that they actually have to do, rather than the raw count values. E.g. Polgar M3s show '[exercises done] / 744' instead of '3718 + [exercises done] / 4462'. This fixes an issue where a task with no progress would always show 0 completed, but adding any progress at all would show 'startCount + progress' completed, and is semantically more consistent with the explanatory text.
change1
- The first entry in the progress history for a task now shows the count the user did for that entry, so e.g. if the user updates Polgar M3s to 3728 for their first entry, the progress history entry will state that they did 10 exercises instead of 3728 as it did previously.
change2

## Type of change

*   [x] Bug fix (non-breaking change which fixes an issue)
*   [ ] New feature (non-breaking change which adds functionality)
*   [ ] Documentation update

## Testing


frontend/src/database/requirement.ts
```diff
@@ -294,7 +304,7 @@ function getYearlyCount({
     const cutoffDate = new Date();
     cutoffDate.setDate(cutoffDate.getDate() - 365);
     const cutoff = cutoffDate.toISOString();
-    let count = 0;
+    let count = requirement.startCount || 0;
```
romyers: I'm not sure what the desired behavior is here. AFAIK there are no yearly tasks with a startCount, and AFAIK it's not possible to make a yearly CustomTask. If, say, someone added a yearly task to do problems x through y from a tactics book, though, I expect we'd want to start at startCount.

romyers:
How feasible is it to fix up the first entry in everyone's progress history in the production database? I recently found that my changes targeting getCurrentCount cause the progress to be rebuilt incorrectly after the timeline is edited. If we could go through and adjust all the first timeline entries in the database, that would be the cleanest fix.

Otherwise, I can think of a couple much messier solutions:
- keep track of exactly when these changes are committed to main and handle timeline entries differently depending on whether they were created before or after that date
- add an attribute to timelineEntries in the database that indicates they were created after this change (e.g. a version number) and handle timeline entries differently depending on that attribute.
- add an alternative to getCurrentCount that always starts from zero, and use that when dealing with the progressHistory (thereby giving up on fixing the inconsistent way progress is presented in the history).

For now, I'll work on the first messy solution.

frontend/src/components/profile/trainingPlan/ProgressHistory.tsx
```diff
@@ -364,7 +364,12 @@ function getTimelineUpdate(
         progress.minutesSpent[item.cohort] =
             (progress.minutesSpent[item.cohort] ?? 0) + minutesSpent;
 
-        const previousCount = progress.counts[cohort] ?? 0;
+        // If a previous count was specified, use that. Otherwise this is the first entry in the timeline.
+        // We want to use requirement.startCount || 0 if the entry version postdates support for this (that is,
+        // if the entry version is truthy at all), and fall back to 0 otherwise.
+        const previousCount =
+            progress.counts[cohort] ?? (item.entry.versionNumber ? requirement.startCount || 0 : 0);
```
romyers: Initial TimelineEntries created before my changes don't behave well when rebuilding the user's progress upon timeline edits, so I added a version number to timeline entries, and fall back to the old behavior when the version number isn't present.

frontend/src/components/profile/trainingPlan/ProgressHistory.tsx
```diff
@@ -434,12 +434,15 @@ export function useProgressHistoryEditor({
...
+        // Older timeline entries in the database may have previousCount < startCount.
+        // It's important to make sure that count will be newCount - startCount
+        // in these cases, so we take Math.max(previousCount, startCount || 0).
         return entries
             .filter((t) => t.requirementId === requirement?.id)
             .sort((a, b) => (a.date || a.createdAt).localeCompare(b.date || b.createdAt))
             .map((t, idx) => ({
                 date: DateTime.fromISO(t.date || t.createdAt),
-                count: `${t.newCount - t.previousCount}`,
+                count: `${t.newCount - Math.max(t.previousCount, requirement?.startCount || 0)}`,
```
romyers: Addresses timeline entries that predate these changes, which might have previousCount = 0 regardless of the value of startCount. We want count to be newCount - startCount in these cases.

PR #2062, opened 2026-03-06
feat: re-filter opening tree in place instead of clearing and reloading

### LINKED ISSUE #2019: feat: Improve speed of repertoire spy
My sense is that this is a very difficult problem, but still one we'd like to have a ticket for. Atm it's mostly not functioning on any search with many games - which most players will have. And when it's searching for a long time it will crash.

jdsalaz:
## Summary

- Add `isLoaded` state flag to `PlayerOpeningTreeProvider` to track whether the opening tree has been loaded
- Replace direct `openingTree.current` ref checks in UI components with the new `isLoaded` state
- Allow `onClear` to reset load state without clearing the tree ref, enabling re-filtering in place

Closes #2019

## Test plan

- [ ] Load the player opening tree and verify it displays correctly
- [ ] Change filter settings and verify the tree re-filters in place without clearing and reloading
- [ ] Verify the "Load Games" button shows/hides appropriately based on load state
- [ ] Verify the filters accordion expands/collapses correctly

### DIFF
--- a/backend/liveClassService/meetingInfo.ts
+++ b/backend/liveClassService/meetingInfo.ts
@@ -77,7 +77,7 @@ export const MEETING_INFO: Record<string, Record<string, MeetingInfo>> = {
 };
 
 function init() {
-    if (process.env.CI === 'true') {
+    if (process.env.CI === 'true' || process.env.VITEST === 'true') {
         return;
     }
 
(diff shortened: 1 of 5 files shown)


jdsalaz:
You're right on both points — thanks for the feedback.

I took a closer look and confirmed that filters already apply in place via `getPosition()` / `matchesFilter()` without reloading. And the `onClear` → `onLoad` flow still calls `loader.load()` which re-fetches everything from the API, so the PR doesn't actually change the reload behavior.

I'm working with some new AI tools and still learning how best to guide them, which led to this misguided attempt at #2019. Closing this one out.

PR #2077, opened 2026-03-08
feat(#2019): add Go game clients and OpeningTree data structure

jdsalaz:
## Summary

- Add Chess.com and Lichess game fetching clients in Go with parallel archive fetching, iterator-based APIs, and a common Game model
- Add OpeningTree data structure for position indexing with merge support, stats calculation, and date/color filtering
- First PR in a stack addressing #2019 — moves opening tree computation from the browser to a backend Lambda

## Related Issues

Addresses #2019

## Type of change

## Testing

* [x] It was not necessary to add tests due to backend-only Go code with Go unit tests included

## Stack

1. **This PR** — Backend core (data structures + game clients)
2. #2078 — Backend service (Lambda handler, API, pagination)
3. #2079 — Frontend integration (Web Worker → backend API)

### DIFF
--- a/backend/openingTreeService/openingtree/bench_test.go
+++ b/backend/openingTreeService/openingtree/bench_test.go
@@ -30,7 +30,7 @@ func TestGoldenFile(t *testing.T) {
 			Result: extractResult(pgn),
 			PGN:    pgn,
 		}
-		_, _ = tree.IndexGame(g)
+		tree.IndexGame(g)
 	}
 
 	got, err := json.MarshalIndent(tree, "", "  ")
@@ -74,7 +74,7 @@ func TestCorrectnessInvariants(t *testing.T) {
 			Result: extractResult(pgn),
 			PGN:    pgn,
 		}
-		_, _ = tree.IndexGame(g)
+		tree.IndexGame(g)
 	}
 
 	t.Run("game_count", func(t *testing.T) {
@@ -188,7 +188,7 @@ func BenchmarkIndex100Games(b *testing.B) {
 	for b.Loop() {
 		tree := New()
 		for _, g := range pool {
-			_, _ = tree.IndexGame(g)
+			tree.IndexGame(g)
 		}
 	}
 }
@@ -199,7 +199,7 @@ func BenchmarkIndex500Games(b *testing.B) {
 	for b.Loop() {
 		tree := New()
 		for _, g := range pool {
-			_, _ = tree.IndexGame(g)
+			tree.IndexGame(g)
 		}
 	}
 }
@@ -210,7 +210,7 @@ func BenchmarkIndex1000Games(b *testing.B) {
 	for b.Loop() {
 		tree := New()
 		for _, g := range pool {
-			_, _ = tree.IndexGame(g)
+			tree.IndexGame(g)
 		}
 	}
 }
@@ -221,7 +221,7 @@ func BenchmarkIndex5000Games(b *testing.B) {
 	for b.Loop() {
 		tree := New()
 		for _, g := range pool {
-			_, _ = tree.IndexGame(g)
+			tree.IndexGame(g)
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


backend/go.mod
```diff
@@ -5,6 +5,7 @@ go 1.25.0
 require (
 	github.com/aws/aws-lambda-go v1.53.0
 	github.com/davecgh/go-spew v1.1.1
+	github.com/notnil/chess v1.10.0
```
jackstenglein: This repo is archived and the README recommends to use https://github.com/corentings/chess instead. Is there a specific reason to use this older one instead?

backend/openingTreeService/chesscom/convert_test.go
```diff
@@ -0,0 +1,107 @@
...
+		White:     Player{Username: "Alice", Rating: 1500, Result: PlayerResultWin},
+		Black:     Player{Username: "Bob", Rating: 1450, Result: PlayerResultCheckmated},
+	}
+
+	common, err := ToGame(g, "alice")
+	if err != nil {
+		t.Fatalf("ToGame error: %v", err)
+	}
+
+	if common.Source != game.SourceChessCom {
```
jackstenglein: This test (and probably many of the others) could be simplified by using [cmp.Diff](https://pkg.go.dev/github.com/google/go-cmp/cmp#Diff)

backend/openingTreeService/game/game.go
```diff
@@ -0,0 +1,59 @@
...
+	Rated         bool       `json:"rated"`
+	URL           string     `json:"url"`
+	Source        SourceType `json:"source"`
+	EndTime       time.Time  `json:"endTime,omitempty"`
+
+	// ArchiveComplete signals that the preceding batch of games from a
+	// Chess.com monthly archive has been fully yielded. EndTime is set to
+	// the first instant of the following month so that FilterArchives with
+	// since=EndTime will exclude the completed archive on resume.
+	ArchiveComplete bool `json:"-"`
```
jackstenglein: It seems weird to me that this is a platform-agnostic representation of a game but then we include a chess.com specific field. I think we should move this out of this struct and define it in the chesscom.go file

backend/openingTreeService/chesscom/chesscom.go
```diff
@@ -281,22 +290,19 @@ func (c *Client) Games(ctx context.Context, username string, since, until time.T
...
-					return
-				}
+				batch = append(batch, cg)
 			}
 
-			// Yield an archive-complete sentinel so the consumer knows
-			// all games from this monthly archive have been delivered.
-			// EndTime is set to the start of the next month so that on
-			// resume FilterArchives(since=EndTime) excludes this archive.
+			// Compute the next-month boundary for cursor updates.
```
jackstenglein: I think the caller should just track the end time of the last returned/indexed game and pass that back as the since parameter in the next call. I foresee an issue with the current approach where we are halfway through the current month and fetch its archive and then return the next month boundary. The caller saves this and passes it back the next day, and we lose indexing all games from the second half of the month.

backend/openingTreeService/openingtree/bench_test.go
```diff
@@ -0,0 +1,341 @@
...
+func TestGoldenFile(t *testing.T) {
+	tree := New()
+	games := splitPGNGames(string(samplePGN))
+	for i, pgn := range games {
+		g := &game.Game{
+			URL:    fmt.Sprintf("https://example.com/game/%d", i),
+			Result: extractResult(pgn),
+			PGN:    pgn,
+		}
+		_, _ = tree.IndexGame(g)
```
jackstenglein: nit: This can just be `tree.IndexGame(g)` without the underscores and equals

backend/openingTreeService/game/game.go
```diff
@@ -0,0 +1,59 @@
...
+// game representations for use by the opening tree and other services.
+package game
+
+import "time"
+
+// SourceType identifies which platform a game originated from.
+type SourceType string
+
+const (
+	SourceChessCom SourceType = "chesscom"
```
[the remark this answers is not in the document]
jdsalaz: Fixed.

PR #2079, opened 2026-03-08
feat(#2019): replace Web Worker with backend API for player opening tree

jdsalaz:
## Summary

- Replace client-side Web Worker opening tree computation with calls to the new openingTreeService Lambda
- Display source errors from backend in the player opening tree UI
- Update Playwright tests for the new backend-driven flow

## Related Issues

Addresses #2019

## Type of change

* [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Testing

## Demo

The user experience is the same — click "Load Games" in the player opening explorer. The difference is computation now happens server-side, so it no longer crashes or times out for players with many games.

## Stack

> **Merge in order** — each PR builds on the previous one.

1. #2077 — Backend core (data structures + game clients)
2. #2078 — Backend service (Lambda handler, API, pagination)
3. **This PR** — Frontend integration (Web Worker → backend API) — merge after #2078


frontend/src/api/explorerApi.tsx
```diff
@@ -86,6 +86,101 @@ export function followPosition(idToken: string, request: FollowPositionRequest)
...
+}
+
+/**
+ * Builds a player opening tree on the backend.
+ * @param idToken The id token of the current signed-in user.
+ * @param sources The player sources to build the tree from.
+ * @returns The serialized opening tree.
+ */
+export function buildPlayerOpeningTree(
+    idToken: string,
```
jackstenglein, line 159: This parameter is unnecessary. axiosService will handle automatically including this if the request URL does not start with `/public/`

jackstenglein (changes requested):
This is slower than our current implementation

PR #2090, opened 2026-03-11
feat(#2000): improve Repertoire Spy discoverability

### LINKED ISSUE #2000: feat: Make repertoire spy easier to find

Kaya-vv:
## Summary

The Repertoire Spy feature is buried 3 levels deep (Analysis Board > Explorer tab > Player tab), making it hard to discover. This PR adds a direct link in the navbar Games dropdown and renames the "Player" tab to "Repertoire Spy" for clarity. A URL deep-link (`?explorer=player`) is also supported so the navbar link can open the correct tab automatically.

## Related Issues

Fixes #2000

## Type of change

## Testing

Two Playwright tests added:
- `explorer.spec.ts`: verifies `?explorer=player` URL param selects the Repertoire Spy tab
- `navbar.spec.ts`: verifies Games dropdown contains Repertoire Spy and navigates correctly

## Demo
Navbar change:
repertoire spy menu

Label change:
repertoire spy label

### DIFF
--- a/frontend/src/board/pgn/explorer/Explorer.tsx
+++ b/frontend/src/board/pgn/explorer/Explorer.tsx
@@ -5,7 +5,7 @@ import CloudIcon from '@mui/icons-material/Cloud';
 import { TabContext } from '@mui/lab';
 import { Box, CardContent, Tab, Tabs } from '@mui/material';
 import { useSearchParams } from 'next/navigation';
-import { useCallback, useEffect, useState } from 'react';
+import { useCallback, useEffect, useRef, useState } from 'react';
 import { SiLichess } from 'react-icons/si';
 import { useLocalStorage } from 'usehooks-ts';
 import { usePosition } from '../../../api/cache/positions';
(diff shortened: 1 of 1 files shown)


frontend/src/board/pgn/explorer/Explorer.tsx
```diff
@@ -36,7 +37,19 @@ export enum ExplorerDatabaseType {
...
+
+    useEffect(() => {
+        const explorer = searchParams.get('explorer');
+        if (
+            explorer &&
+            Object.values(ExplorerDatabaseType).includes(explorer as ExplorerDatabaseType)
+        ) {
+            setTab(explorer as ExplorerDatabaseType);
+        }
+    }, []); // eslint-disable-line react-hooks/exhaustive-deps
```
jackstenglein: In general for this project, I highly discourage eslint-disable comments. I think there are a number of ways to achieve only-once behavior without disabling the linter. You can pick your favorite from these:

- Extract the body of the useEffect to a useEffectEvent and then call that in useEffect
- Clear the searchParams after setting the tab value
- Add a ref that tracks whether we have already performed this behavior, and set it to true when we set the tab. Then only set the tab if the ref is false.

jackstenglein (changes requested):
Please fix the eslint issue and otherwise this looks good

Issue #2109, opened 2026-03-13
Conserve space on training screen TECHGEN#427


jackstenglein:
**💬 Content:** 
  > I love that we can now minimize "today" if we want to primarily see "this week" or other way around.  But the clickable group tabs still take up valuable real estate on the screen.  Any way that of something is minimized that header can be moved adjacent to the other header or beneath the full training program group?

👤 Author name: mattchess
[Discord URL](https://discord.com/channels/951958534113886238/1026952329804644372/1482130525228372028)

PR #2112, opened 2026-03-14
Feat/2019 training screen space

bourama1:
## Summary

When a training section ("Today" or "This Week") is collapsed, its header now shrinks to half-width and flows inline next to the other section header, rather than occupying a full row. Expanded sections continue to take full width. Collapse state is persisted via localStorage. On mobile, sections always stack full-width regardless of state.

* Lifted expanded state for both daily and weekly sections up to TrainingPlanTab so layout can react to both states simultaneously
* Replaced fixed-width wrappers with flexWrap containers — collapsed sections shrink to calc(50% - 8px) and sit side-by-side, expanded sections take full width
* Converted expanded/setExpanded from local state to props in DailyTrainingPlan and WeeklyTrainingPlan
* Applied the same responsive reflow to FullTrainingPlanSection accordions, with disableGutters and removed default divider for a cleaner collapsed appearance
* Made AccordionSummary headers sticky (offset by navbar height) so titles stay visible when scrolling long expanded sections

## Related Issues

Fixes [# (2019)](https://github.com/jackstenglein/chess-dojo/issues/2109)

## Type of change

*   [ ] Documentation update

## Testing

* [x] It was not necessary to add tests due to not being sure if those changes are what was mean by issue 

## Demo

thorium_tp8kMgbquW

### DIFF
--- a/frontend/src/components/profile/trainingPlan/TrainingPlanTab.tsx
+++ b/frontend/src/components/profile/trainingPlan/TrainingPlanTab.tsx
@@ -40,7 +40,10 @@ export function TrainingPlanTab({ user }: { user: User }) {
                             width: dailyExpanded ? 1 : { xs: 1, sm: 'calc(50% - 8px)' },
                         }}
                     >
-                        <DailyTrainingPlan expanded={dailyExpanded} setExpanded={setDailyExpanded} />
+                        <DailyTrainingPlan
+                            expanded={dailyExpanded}
+                            setExpanded={setDailyExpanded}
+                        />
                     </Box>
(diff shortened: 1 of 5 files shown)


jackstenglein:
@hellokostya @jessekraai What do you think of the screenshot shared above? I think it looks a bit strange personally, but I'm not sure if it's just because I am used to the current layout

hellokostya:
@jackstenglein Not sure about this view either, want to maybe think about the overall design

jessekraai:
Not sure, would want to play with it

PR #2125, opened 2026-03-16
feat(#2073): add square color drill

### LINKED ISSUE #2073: What color is this square drill master
Should be ez! We ask a square and the student must answer the color. Eg. g7 answer: black.

The challenge is to then create a rating system around it. And I think the pro move is for now all we need to do is record how long it takes for the players to answer and their accuracy scores. We tell our beta testers to answer asap and then we construct a rating system based on their accuracy and response times.

This will then be a tactics requirement for the U/1000 cohorts.

Kaya-vv:
## Summary

Adds a "What color is this square?" drill where users are shown a random square (e.g. g7) and must answer whether it's black or white. Tracks accuracy, response time, and best streak per session. Users can stop at any time and view a summary of their results. Sessions are persisted to DynamoDB for future use in a rating system.

## Related Issues

Fixes #2073

## Type of change

## Testing

## Demo

**Tests page with new Square Color Drill card:**
puzzle nav

**Start screen with instructions:**
test start

**Drill in progress, answer with buttons or keyboard (W/B):**
questionnaire
**Session summary with accuracy, timing, and question history:**
session complete

### DIFF
--- a/common/src/squareColors/squareColor.test.ts
+++ b/common/src/squareColors/squareColor.test.ts
@@ -1,8 +1,5 @@
-import {
-    getRandomSquare,
-    getSquareColor,
-} from '@jackstenglein/chess-dojo-common/src/squareColors/squareColor';
 import { describe, expect, it } from 'vitest';
+import { getRandomSquare, getSquareColor } from './squareColor';
 
 describe('getSquareColor', () => {
     it('returns black for a1', () => {
--- a/frontend/src/api/Api.tsx
+++ b/frontend/src/api/Api.tsx
@@ -30,7 +30,6 @@ import {
     RoundRobinSubmitGameRequest,
     RoundRobinWithdrawRequest,
 } from '@jackstenglein/chess-dojo-common/src/roundRobin/api';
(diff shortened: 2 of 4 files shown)


common/src/squareColors/squareColor.ts
```diff
@@ -0,0 +1,25 @@
...
+ * A square is dark if (fileIndex + rankIndex) is even (0-indexed).
+ */
+export function getSquareColor(square: string): 'black' | 'white' {
+    const file = FILES.indexOf(square[0]);
+    const rank = parseInt(square[1], 10) - 1;
+    return (file + rank) % 2 === 0 ? 'black' : 'white';
+}
+
+/**
+ * Returns a random square name from a1-h8.
```
jalpp: Same here

jalpp (approved):
LGTM @jackstenglein you can double check

jackstenglein:
@Kaya-vv Thanks, this is a great first version! The only issue I notice is that we only save the session if the user clicks stop. If they just leave the page, then we don't actually save any of their results. Not a big deal for this version, but I opened #2144 for that improvement if you want to take that one up as well.

PR #2127, opened 2026-03-16
feat: add mini cohort leaderboard to training plan

### LINKED ISSUE #2104: Add scoreboard underneath Full Training Plan on training page
On the training page, underneath the full training plan there is currently an empty space, maybe we can put a reduced version of the cohort scoreboard and link to the full scoreboard, similar to the newsfeed panel on the left

H4ROOON:
## Summary
new MiniScoreboard component that sits below the training plan to fill the empty space kostya flagged.
wired it to the parent cohort state so the leaderboard updates automatically when the user switches rating ranges. reused API.getScoreboard with a typescript type guard to filter out graduations and summaries, showing only the top 5 users by dojo score.
handles loading, empty states, and API failures so nothing breaks. includes a link to the full scoreboard.

## Demo

Screenshot 2026-03-22 174648

closes #2104

### DIFF
--- a/frontend/src/components/profile/trainingPlan/MiniScoreboard.tsx
+++ b/frontend/src/components/profile/trainingPlan/MiniScoreboard.tsx
@@ -1,0 +1,250 @@
+import { useApi } from '@/api/Api';
+import { useRequest } from '@/api/Request';
+import { useAuth } from '@/auth/Auth';
+import { Link } from '@/components/navigation/Link';
+import { User } from '@/database/user';
+import Avatar from '@/profile/Avatar';
+import CohortIcon from '@/scoreboard/CohortIcon';
+import { ScoreboardRow } from '@/scoreboard/scoreboardData';
+import WorkspacePremiumIcon from '@mui/icons-material/WorkspacePremium';
+import {
+    Button,
+    Card,
+    CardContent,
+    CircularProgress,
+    Divider,
+    MenuItem,
+    Stack,
+    TextField,
+    Typography,
+} from '@mui/material';
+import { useEffect, useState } from 'react';
+
+/**
+ * Type guard to filter the scoreboard API response.
+ * Ensures we only rank actual player profiles, not graduation events.
+ * @param row - A single data row from the scoreboard API response.
+ * @returns True if the row is a valid User profile.
+ */
+const isUser = (row: ScoreboardRow): row is User => {
+    return 'username' in row && 'displayName' in row && 'progress' in row;
+};
+
+/**
+ * Rounds the user's total dojo score to 2 decimal places.
+ * @param user - The user whose score to calculate.
+ * @returns The rounded dojo score.
+ */
+const getScore = (user: User): number => {
+    return Math.round((user.totalDojoScore ?? 0) * 100) / 100;
+};
+
+/**
+ * Gets the user's total training time in minutes.
+ * @param user - The user object.
+ * @returns Total minutes spent.
+ */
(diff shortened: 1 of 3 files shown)


H4ROOON:
hey @hellokostya have a look at the demo image and let me know if this is something you had in mind.

frontend/src/components/profile/trainingPlan/MiniScoreboard.tsx
```diff
@@ -0,0 +1,148 @@
...
+    CardContent,
+    CircularProgress,
+    Divider,
+    Stack,
+    Typography,
+} from '@mui/material';
+import { useEffect } from 'react';
+
+// the scoreboard API returns a mixed array of users, graduations, and summaries.
+// added this filter to ensure only actual player profiles get ranked.
```
jalpp: can we please follow Jsdoc comments style for functions, what params mean gets lost when looking at this function from outside

frontend/src/components/profile/trainingPlan/MiniScoreboard.tsx
```diff
@@ -0,0 +1,148 @@
...
+const isUser = (row: ScoreboardRow): row is User => {
+    return 'username' in row && 'displayName' in row && 'progress' in row;
+};
+
```
jalpp: Same here please add jsdoc comment

jalpp (approved):

jalpp:
@jackstenglein this PR looks good to me, can double check

frontend/src/components/profile/trainingPlan/full/FullTrainingPlan.tsx
```diff
@@ -322,6 +323,7 @@ export function FullTrainingPlan() {
                         setShowCompleted={setShowCompleted}
                     />
                 ))}
+                <MiniScoreboard cohort={cohort} />
```
jackstenglein: I think it would make more sense to add this to TrainingPlanTab.tsx instead of here

frontend/src/components/profile/trainingPlan/MiniScoreboard.tsx
```diff
@@ -0,0 +1,172 @@
...
+                    data-testid='mini-scoreboard-headers'
+                >
+                    <Typography variant='caption' color='text.secondary' fontWeight='bold'>
+                        Name
+                    </Typography>
+                    <Typography variant='caption' color='text.secondary' fontWeight='bold'>
+                        Score
+                    </Typography>
+                </Stack>
+                {topPlayers.map((player, index) => {
```
jackstenglein: We should also show the current player, even if they are not in the top 5. I think free users won't be returned by the scoreboard API, so you will probably have to calculate the current user's score directly from the user object. That will also have the benefit of updating their score automatically when they update their progress.

jackstenglein:
@hellokostya Do we want to show training time or dojo score? I think most users won't really know how the Dojo score works so it may be better to show training time

hellokostya:
@jackstenglein @H4ROOON could there be enough room for both, with the ability to sort for one or the other?

H4ROOON:
@hellokostya totally doable. cramming both time and score might get too cluttered on mobile i think. we can do a  dropdown/toggle to switch between them. will wait for @jackstenglein call on it.

p.s

ill be afk for eid holidays, will do the requested changes a day after tomorrow.

jackstenglein:
I think a dropdown like we have on the heatmap would work well:

Screenshot 2026-03-21 at 10 14 18 AM

PR #2130, opened 2026-03-18
feat(#2126): discord bot for posting blogs

### LINKED ISSUE #2126: Discord Bot For Posting Blogs
We should have a bot that posts our blogs in Discord, maybe this could be the bot that posts the weekly stream / live class schedule as well.

https://chessdojo.club/blog

Kaya-vv:
## Summary

Automatically posts a Discord embed to the blog channel when a blog post is published. The notification is triggered via an SQS event from the blog service to the notification service, which builds a Discord embed with the blog's title, subtitle (as the author field), description, cover image, and a link to the blog post. The message includes a role ping and a randomized announcement message.

**Note:** I don't have access to the dev Discord server/channels, so `discordBlogChannelId` and `discordBlogRoleId` in config-dev.yml are set to the prod values. These should be updated to dev channel/role IDs before testing on the dev stage.

## Related Issues

Fixes #2126

## Type of change

* [ ] Documentation update

## Testing

* [x] It was not necessary to add tests due to this being a backend-only feature using SQS and the Discord bot, with no existing test infrastructure for the notification service

## Demo

Expected Discord output:

image

### DIFF
--- a/backend/blogService/create.ts
+++ b/backend/blogService/create.ts
@@ -1,10 +1,6 @@
 'use strict';
 
-import {
-    ConditionalCheckFailedException,
-    PutItemCommand,
-    UpdateItemCommand,
-} from '@aws-sdk/client-dynamodb';
+import { ConditionalCheckFailedException, PutItemCommand } from '@aws-sdk/client-dynamodb';
 import { marshall } from '@aws-sdk/util-dynamodb';
 import {
     Blog,
@@ -72,6 +68,7 @@ async function createBlog(request: CreateBlogRequest): Promise<Blog> {
         createdAt: updatedAt,
         updatedAt,
         status,
+        ...(status === BlogStatuses.PUBLISHED && { discordPosted: true }),
     };
 
     try {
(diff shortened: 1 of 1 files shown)


backend/blogService/create.ts
```diff
@@ -89,5 +94,48 @@ async function createBlog(request: CreateBlogRequest): Promise<Blog> {
         throw new ApiError({ statusCode: 500, publicMessage: 'Internal server error', cause: err });
     }
 
+    if (status === BlogStatuses.PUBLISHED) {
+        try {
+            await sendBlogPublishedEventIfFirst(blog);
```
jackstenglein: This is already guaranteed to be a new blog due to the condition expression on the put item command above. I think we can just update the blog item above to have `discordPosted` set to true when the blog is created, and call `sendBlogPublishedEvent` here

jackstenglein: Note: discordPosted should still only be true if the blog is created in published status

PR #2136, opened 2026-03-18
feat(#2135): add Stockfish 18 and Stockfish 18 Lite engine options

### LINKED ISSUE #2135: Replace SF 17 with 18
A nice improvement that shouldn't be so hard.

echo8134:
## Summary

- Adds Stockfish 18 (dual-net, 108 MB) and Stockfish 18 Lite (smallnet, 16 MB) as new engine options using WASM builds from `@lichess-org/stockfish-web`
- SF 18 is now the default engine; SF 17.1 and SF 16.1 descriptions updated to "Previous version"
- The SF 18 big NNUE net is stored gzipped (70 MB) to fit under GitHub's 100 MB file limit and decompressed at runtime using `DecompressionStream`. Decompressed bytes are cached in IndexedDB so decompression only happens on first download.

## Notes for reviewer

- The `.js` files (`sf18.js`, `sf18-smallnet.js`) are vendored from the npm package with two patched locations each (WASM binary path and pthread worker path) - same approach as existing `sf171-79.js`
- Only the big NNUE net (`nn-c288c895ea92.nnue`) is gzipped. Other NNUE files are under 100 MB and served as-is. A future PR could compress all NNUE files for consistency and smaller downloads.
- The `@lichess-org/stockfish-web` package is AGPL-3.0 licensed

## Test plan

- [x] Toggle engine ON - should show "SF 18" and eval lines after NNUE download
- [x] Switch to "Stockfish 18 NNUE Lite" - should reinitialize with smaller download
- [x] Switch to SF 17.1, SF 16, SF 11 - verify they still work
- [x] Verify NNUE caching: refresh page, SF 18 should load faster (IndexedDB cache)
- [x] Verify engine settings (depth, threads, hash, multi-PV) persist across refresh
- [x] Check Network tab shows `nn-c288c895ea92.nnue.gz` (not `.nnue`) being fetched

Closes #2135


echo8134:
image

Should correctly load. Stress-tested with different settings too

PR #2150, opened 2026-03-21
fix(#2101): prevent error wrapping on successful timeline entry insertion

### LINKED ISSUE #2101: Internal server error Tech_BUG#133
**📓 Steps to reproduce:** 
  > Keep getting an internal server error when trying to log tasks on the site. The request still gets through as the tasks do get logged but get the delay and red error message each time
🔨 Platform: chessdojo.club
📱 Device Type: dojo phone app, desktop firefox and brave
🌐 Browser Type: dojo phone app, desktop firefox and brave

👤 Author name: George
[Discord URL](https://discord.com/channels/951958534113886238/1299913797045325957/1482078400788172810)

Dblike:
## Summary

This PR fixes an internal server error that occurs when logging tasks, even though the tasks are successfully saved. The issue was likely caused by the task logging function timing out before completing milestone notifications to Discord. By increasing the function's timeout threshold, the notifications should now complete successfully and users no longer see the error message.

**Changed files:** `backend/user/serverless.yml`

## Related Issues

Fixes #2101

## Type of change

* [x] Bug fix (non-breaking change which fixes an issue)
* [ ] New feature (non-breaking change which adds functionality)
* [ ] Documentation update

## Testing

* [x] It was not necessary to add tests due to the change being a targeted fix with manual verification

### Manual testing steps

- [x] Step 1: Deploy the updated serverless.yml to a test environment using `sls deploy -s <stage>`
- [x] Step 2: Log into chessdojo.club with a test account
- [x] Step 3: Navigate to the training program and select a task to log progress on
- [x] Step 4: Update the task progress (e.g., increment count by 1) and click save
- [x] Step 5: Verify that NO red error message appears and the UI shows success
- [x] Step 6: Verify the task progress was saved by refreshing the page and checking the updated count
- [x] Step 7: Check CloudWatch logs for the updateProgress Lambda to confirm no timeout errors
- [x] Step 8: (Edge case) If testing with a user at exactly 85% completion, verify that sensei DMs are sent without causing a timeout
- [x] Step 9: Test on multiple platforms mentioned in the bug report: mobile app, desktop Firefox, and Brave browser

## Demo

Backend-only change — no frontend demo applicable.

### DIFF
--- a/backend/user/serverless.yml
+++ b/backend/user/serverless.yml
@@ -135,6 +135,7 @@ functions:
 
   updateProgress:
     handler: progress/update/main.go
+    timeout: 10
     events:
       - httpApi:
           path: /user/progress/v2


Dblike:
@jackstenglein Couple notes for potential follow up work:
- Make milestone notifications async — milestone.checkNotification() runs synchronously before the response is sent. It pages through all Discord guild members in batches of 1000 via GetSenseiDiscordIds(), then sends DMs to each sensei. This should be dispatched to SQS or a separate Lambda so the user gets their response immediately after the DB writes succeed. A failed notification shouldn't block or error the user's task logging.

- Reduce response payload — api.Success() marshals the entire User object including Progress (every requirement ever worked on), RatingHistories, OpeningProgress, and Exams. For active users this is very large. The frontend's ProgressUpdater.tsx only needs the updated progress entry and timeline entry — not the full user. Returning a slimmed-down response would reduce serialization time and network transfer.

PR #2160, opened 2026-03-23
fix(#2151): Deleted game still shows up in games list

### LINKED ISSUE #2151: Deleted game still shows up in games list Tech_BUG#433
**📓 Steps to reproduce:** 
  > 1. I added a game (abahl12 v blueguitargambit) and saved it
2. I then made changes to the pgn offline (hiarcs) and copied to the clipboard
3. In the dojo pgn editor, I clicked the gear and the Replace PGN button and pasted my pgn (I scrolled through it to verify it had my changes and pasted the pgn.
4.  The PGN did not update in the game. 
5. I repeated step 3 again (also verifying the pasted pgn was the intended clipboard copy) and submitted.
6. The PGN did not update again so I deleted the game in the dojo editor (I don't recall the response to this step unfortunately) but I think it left me in the Home > My Games page.
7. I then went to my training plan - refreshed the browser and navigated back to my games 
8. The game still showed there but clicking it gave me a 404 Resource Not Found.
9. I then added a new game , pasted the clipboard PGN and saved. It saved correctly.
10. So I have a phantom entry that I cannot get rid of - It is the FIRST entry (surprising to me). So please delete the corrupted (first entry) of albalh12 v blueguitar gambit  (or both entries if needed, I can re-add it). 

So, two bugs I think, 1) could not reload PGN to a saved (but not public) game (actually, I don't recall if I saved the game first or not, sorry about that) but I have reloaded PGN from a non saved game in the past (3/19), and 2) somehow I corrupted the game and it could not be deleted.

Thanks
🔨 Platform: chessdojo.club
📱 Device Type: Desktop
🌐 Browser Type: Chrome 145.0.7632.160 (Official Build) (x86_64) on OSX

👤 Author name: blueguitargambit
[Discord URL](https://discord.com/channels/951958534113886238/1299913797045325957/1485030543694368899)

Dblike:
## Summary

When a game was deleted, the `deleteGames` handler removed it from the games table but left stale references in the user's directory entries. This caused deleted games to still appear in the games list, returning 404 when clicked.

The fix:
- Changes `deleteGames` to return `ALL_OLD` from the DynamoDB delete, read the game's `directories` field, and call `removeDirectoryItems` to clean up each directory reference
- Extracts the delete + cleanup logic into a `deleteGameAndCleanup` function
- Adds the required IAM permissions (`GetItem` + `UpdateItem` on directories table, `PartiQLUpdate` on games table)
- Includes a one-time backfill script (`cleanupPhantomGames.ts`) to remove existing phantom entries on prod

## Related Issues

Fixes #2151

## Type of change

* [x] Bug fix (non-breaking change which fixes an issue)
* [ ] New feature (non-breaking change which adds functionality)
* [ ] Documentation update

## Testing

* [x] It was not necessary to add tests due to the change being a targeted fix with manual verification

### Manual testing steps

- [x] Open a game you own, click the gear icon, select "Replace PGN"
- [x] Paste a modified PGN (e.g., add a comment) and submit
- [x] Verify the game view shows the updated PGN immediately (no stale data)
- [x] Navigate away and back — updated PGN should persist
- [x] Open a game you own, delete it from the game detail page
- [x] Verify the games list does not show a phantom entry for the deleted game
- [x] Verify clicking through directories shows no 404 entries for the deleted game
- [x] Import a new game — verify `?firstLoad=true` preflight dialog still works normally


backend/pgnService/game/delete.ts
```diff
@@ -76,6 +78,24 @@ async function deleteGames({
     for (let i = 0; i < results.length; i++) {
         if (results[i].status === 'fulfilled') {
             deleted.push(request[i]);
+
+            // Clean up directory references for the deleted game
+            const result = results[i] as PromiseFulfilledResult<{ Attributes?: Record<string, unknown> }>;
```
jackstenglein: I think it would make more sense to extract the dynamo.send call up above into a new async function and add the removeDirectoryItems call into that function as well for each game. That would keep things a bit cleaner

frontend/src/games/view/GamePage.tsx
```diff
@@ -54,14 +54,17 @@ const GamePage = ({ cohort: initialCohort, id: initialId }: { cohort: string; id
     const reset = request.reset;
     useEffect(() => {
         if (cohort && id) {
+            if (firstLoad) {
```
jackstenglein: I'm not sure I understand the problem this solves from your PR description. Could you add a screen recording of the bug in prod or include some steps to reproduce it? This seems unrelated to #2151 so I'm a bit confused why we need to make this change

frontend/src/games/list/ListGamesPage.tsx
```diff
@@ -119,6 +119,7 @@ const ListGamesPage = () => {
                             .filter((g) => !!g)}
                         onClose={contextMenu.close}
                         position={contextMenu.position}
+                        setGames={setGames}
```
jackstenglein: I'm not sure I understand the problem this solves from your PR description. Could you add a screen recording of the bug in prod or include some steps to reproduce it? This seems unrelated to #2151 so I'm a bit confused why we need to make this change

jackstenglein (changes requested):

frontend/src/games/view/GamePage.tsx
Dblike: This change is related to a separate issue where replacing a game's PGN via the editor shows stale data because the module-level gameCache is never invalidated. I've reverted it from this PR and split it into a separate branch (fix/2151-replace-pgn-cache) to keep this PR focused on the phantom deleted games fix.

frontend/src/games/list/ListGamesPage.tsx
Dblike: Reverted this change. Realized I had a permissions issue on the lambda call and this changed was unnecessary.

jackstenglein (approved):

PR #2166, opened 2026-03-25
Activity piechart breakdown

### LINKED ISSUE #2097: Hover over activity pie graph
At the moment you have to know to click on a section of the pie graph to see the breakdown of activities. Scrytop right called this "deeply undiscoverable"! Improvement: we should be able to see what the activities are when we hover over the different sections of the pie.

JChristensen97-sec:
## Summary

Updated both Activity pie charts in /profile?view=activity so hovering a top-level slice now shows its subcategory breakdown directly in the tooltip, without requiring drilldown. Tooltip formatting was also improved by emphasizing the parent category and sorting subcategories in descending order.

## Related Issues

Fixes  #2097 

## Type of change

*   [x] Bug fix (non-breaking change which fixes an issue)
*   [ ] New feature (non-breaking change which adds functionality)
*   [ ] Documentation update

## Testing

* [x] It was not necessary to add tests due to being a minor UI enhancement with no underlying data logic changes.

## Demo

If this PR makes changes to the frontend, please include a screenshot or screen recording of the changes.
Changes the time cohort to look like this:
image

Changes the score cohort to look like this: 
image


frontend/src/components/profile/activity/ActivityPieChart.tsx
```diff
@@ -15,28 +15,12 @@ import { UseTimelineResponse } from './useTimeline';
  * @param entry The PieChartData entry to get the tooltip for.
  * @returns The string for the tooltip.
```
jalpp: why are comments still here if the function was removed?

frontend/src/components/profile/activity/ActivityPieChart.tsx
```diff
@@ -15,28 +15,12 @@ import { UseTimelineResponse } from './useTimeline';
...
-    }
-    const score = Math.round(entry.value * 100) / 100;
-    if (entry.count) {
-        return `${entry.name} - Count: ${entry.count}, Score: ${score}`;
-    }
-    return `${entry.name} - ${score}`;
-}
 
 /**
  * Returns a string for the time chart tooltip.
```
jalpp: same here why is the comment here?

frontend/src/components/profile/activity/ActivityPieChart.tsx
```diff
@@ -126,6 +110,90 @@ const ActivityPieChart: React.FC<ActivityPieChartProps> = ({ user, timeline }) =
         );
     }, [user, cohorts, timeframe, timeline.entries, timeChartCategory, requirements]);
```
jalpp, line 121: we should add jsdoc to this function also

frontend/src/components/profile/activity/ActivityPieChart.tsx
```diff
@@ -126,6 +110,90 @@ const ActivityPieChart: React.FC<ActivityPieChartProps> = ({ user, timeline }) =
...
+                {breakdown.map((item) => (
+                    <div key={item.name} style={{ paddingLeft: 8 }}>
+                        {item.name} - {getTimeDisplay(item.value)}
+                    </div>
+                ))}
+            </div>
+        );
+    };
+
+    const getScoreChartTooltip = (entry?: PieChartData) => {
```
jalpp, line 209: same here

jalpp (changes requested):

JChristensen97-sec:
Thanks for the catch @jalpp .  That was a sloppy miss on my part.  I cleaned up the stale comments and added JSDoc for the new tooltip helpers, so this should be resolved now.

jalpp:
> Thanks for the catch @jalpp . That was a sloppy miss on my part. I cleaned up the stale comments and added JSDoc for the new tooltip helpers, so this should be resolved now.

no worries, you should hit resolve convo on Github once you made requested changes, there are still few things I will add now.

frontend/src/components/profile/activity/ActivityPieChart.tsx
```diff
@@ -126,6 +98,110 @@ const ActivityPieChart: React.FC<ActivityPieChartProps> = ({ user, timeline }) =
...
+                requirements,
+            ),
+        ].sort((a, b) => b.value - a.value);
+
+        if (!breakdown.length) {
+            return base;
+        }
+
+        return (
+            <div>
```
jalpp: can we stick to using MUI's box instead of div and use sx styling. It makes easier to read

frontend/src/components/profile/activity/ActivityPieChart.tsx
```diff
@@ -126,6 +98,110 @@ const ActivityPieChart: React.FC<ActivityPieChartProps> = ({ user, timeline }) =
...
+                requirements,
+            ),
+        ].sort((a, b) => b.value - a.value);
+
+        if (!breakdown.length) {
+            return <div>{base}</div>;
+        }
+
+        return (
+            <div>
```
jalpp: same here use Box and sx

frontend/src/components/profile/activity/ActivityPieChart.tsx
```diff
@@ -126,6 +98,110 @@ const ActivityPieChart: React.FC<ActivityPieChartProps> = ({ user, timeline }) =
...
+                <div style={{ fontWeight: 700, marginBottom: 4 }}>{base}</div>
+                {breakdown.map((item) => {
+                    const childScore = Math.round(item.value * 100) / 100;
+                    return (
+                        <div key={item.name} style={{ paddingLeft: 8 }}>
+                            {item.count
+                                ? `${item.name} - Count: ${item.count}, Score: ${childScore}`
+                                : `${item.name} - ${childScore}`}
+                        </div>
```
jalpp: for both score and time, for parent category we should add icon related to that category eg: swords for games etc which you can find in icon.tsx or the training plan view. 2nd we should bold the time/scores so its easier to read and also provide % break down of overall time vs sub time, same for score. so if games has 20 hr logged and within games there was 8 hr of test we should show the ratio in percent. Since we reading from pie chart before tool tip, user would expect to see tool tip containing precents. Or even better if you can show sub pie chart based on logic above. The current way is doable, but I prefer sub pie chart from UI/UX point of view.

jalpp (changes requested):

frontend/src/components/profile/activity/ActivityPieChart.tsx
jackstenglein: We definitely should not show a smaller pie chart inside the tooltip. That would be way too cramped and a bad UX. I'm ambivalent about including the percentages inside the tooltip text. My guess is that will be a bit too much text on the score pie chart, but I'm not sure

jackstenglein: I do agree adding the icons/bolding the values could be a nice touch

JChristensen97-sec: Yeah, I was about to add that I think a second pie chart would just look over-congested.  
I can work on the formatting issues.
Perhaps, I could also add something like "click for more details" at the very bottom to make the sub-category pie chart more intuitive to find as a different solution.

jalpp: yeah that was my suggestion, my point was we should point the user to that pie chart or do something that bolds the info and makes it easy to understand. Right now its very paragraph like and not formatted as a user I wouldn't even read that tool tip if there nothing bold or some eye catching thing @JChristensen97-sec @jackstenglein

JChristensen97-sec:
I updated the tooltip formatting to use Box and sx, added icons for the main categories, and adjusted the typography so the parent category is more prominent and the subcategories are easier to scan.

I held off on adding percentages for now, since the tooltip was already starting to feel congested and I did not see a clean way to include them without hurting readability.

I attached a few screenshots showing the updated layout.
  
image
image

jalpp (approved):
LGTM

jackstenglein (approved):

PR #2170, opened 2026-03-25
fix(#2165): improvements to game annotation clock editor

### LINKED ISSUE #2165: Improvements to Game Annotation
1) If you refresh the page without saving the analysis first you are toast. Can we fix this?
2) When you are entering times you have to put in the hour time, then use the mouse to click into the minutes window. Would be great if we could use the right arrow for this instead, timesaver!
3) The hour box should become highlighted in red (ie a mistake has been made dear user!) if the hours are greater than the last move. This helps prevents a typo from wrecking the time graph and rating.
4) If you are under an hour the cursor should automatically be in the minutes box

Dblike:
## Summary

Improves the game annotation clock editor based on user feedback. Adds arrow key navigation between clock fields, visual validation when hours exceed the previous move, auto-focus to the minutes field for sub-1-hour time controls, and restores the browser warning when refreshing with unsaved analysis.

## Related Issues

Fixes #2165

## Type of change

* [ ] Documentation update

## Testing

* [x] It was not necessary to add tests due to UI interaction changes best verified through manual testing

### Manual testing steps

- [x] Import a game and navigate to the Clocks tab in the analysis board
- [x] Switch to the three-field clock format (Hours/Minutes/Seconds) in editor settings
- [x] Click into the Hours field and press Right Arrow — cursor should move to Minutes, then to Seconds
- [x] From Seconds, press Left Arrow — cursor should move back through Minutes to Hours
- [x] Enter a time where hours exceed the previous move's hours — the Hours field should highlight red
- [x] For a game with a sub-1-hour time control, click into a clock row — cursor should auto-focus to Minutes on first click
- [x] Click the Hours field again — cursor should stay on Hours (redirect only happens once)
- [x] Make changes to the analysis, then try to refresh the browser — a "Changes may not be saved" warning should appear
- [x] Cancel the refresh and verify your changes are still intact

### Notes

- The browser only allows its native "Leave site?" prompt on refresh — the custom Save/Delete dialog is only possible for in-app navigation due to browser security restrictions. The native prompt still prevents accidental data loss.
- Focus redirects from Hours to Minutes on first click when the previous move's clock is under 1 hour. Clicking Hours a second time stays on Hours, and Left Arrow from Minutes also reaches Hours. This is per-player — if one player is under an hour but the other isn't, each row behaves appropriately.
- The hours validation (red highlight) assumes a live game where clock time decreases. For correspondence games where time can increase between moves, the red highlight may be a false positive. This is a visual hint only — it does not prevent saving.

## Demo

image

### DIFF
--- a/frontend/src/app/(scoreboard)/games/analysis/AnalysisBoard.tsx
+++ b/frontend/src/app/(scoreboard)/games/analysis/AnalysisBoard.tsx
@@ -20,7 +20,7 @@ import {
 } from '@jackstenglein/chess-dojo-common/src/database/game';
 import { Button, Dialog, DialogActions, DialogContent, DialogTitle } from '@mui/material';
 import { useNavigationGuard } from 'next-navigation-guard';
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
 
 const gameUrlRegex = /^\/games\/.*\/.*/;
 
@@ -48,6 +48,17 @@ export default function AnalysisBoard() {
         onSubmit,
         request,
     } = useUnsavedGame(chess);
+
+    useEffect(() => {
+        const handleBeforeUnload = (event: BeforeUnloadEvent) => {
+            event.preventDefault();
+            // Modern browsers ignore custom messages, but returnValue must be set
+            event.returnValue = '';
+        };
+
+        window.addEventListener('beforeunload', handleBeforeUnload);
+        return () => window.removeEventListener('beforeunload', handleBeforeUnload);
+    }, []);
 
     if (status === AuthStatus.Loading) {
         return <LoadingPage />;
--- a/frontend/src/app/(scoreboard)/games/analysis/AnalysisBoard.tsx
+++ b/frontend/src/app/(scoreboard)/games/analysis/AnalysisBoard.tsx
@@ -20,7 +20,7 @@ import {
 } from '@jackstenglein/chess-dojo-common/src/database/game';
 import { Button, Dialog, DialogActions, DialogContent, DialogTitle } from '@mui/material';
 import { useNavigationGuard } from 'next-navigation-guard';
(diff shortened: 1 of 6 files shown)


frontend/src/app/(scoreboard)/games/analysis/AnalysisBoard.tsx
```diff
@@ -49,6 +49,17 @@ export default function AnalysisBoard() {
         request,
     } = useUnsavedGame(chess);
 
+    useEffect(() => {
```
jackstenglein: I don't think this useEffect is necessary. The production deployment already has a warning if you try to navigate away or refresh with an unsaved analysis:

Image

frontend/src/board/pgn/boardTools/underboard/clock/ClockEditor.tsx
```diff
@@ -54,6 +55,11 @@ const ClockEditor = ({
     const isThreeField = clockFieldFormat === ClockFieldFormat.ThreeField;
 
     for (let i = 0; i < moves.length; i += 2) {
+        // Get previous move's clock for validation (previous black move for white, previous white move for black)
+        const whitePrevMoveSeconds =
```
jackstenglein: This is getting the wrong move to compare against. You're comparing white's clock on the current move to black's clock on the previous move:

Image

I think you need

```typescript
const whitePrevMoveSeconds = i >= 2 ? clockToSeconds(moves[i - 2]?.commentDiag?.clk) : undefined;
const blackPrevMoveSeconds = clockToSeconds(moves[i-1]?.commentDiag?.clk);
```

jackstenglein (changes requested):

jackstenglein (approved):

PR #2171, opened 2026-03-26
feat(#2149): Square drill ratings

### LINKED ISSUE #2149: Square Color Drill enhancement
The new square color drill looks great. We need two things:
1) the user needs to know that they are being timed from the first moment. So we need to tell them that and something like a ready? GO! button needs to be in place
2) are we collecting all the data? This little app is only cool if we can give ppl a rating.
--> How should this rating work? 65% of the score should be accuracy and 35% time.
--> The ceiling should be 1500. In our program we are only going to ask ppl to do it up to 1000. Eg. if you are in the 800-900 cohort we will ask that you achieve a rating of 900. There are two ways to do this, the ez and the hard:
a) the ez: We get several ppl at various rating ranges from 0-1500 to do the test and thereby set benchmarks. This could conceivably avoid the inflation problem that our mate server has.
b) the hard: we gather all the data and continually update the ratings. More work and the inflation problem.

I vote for ez! If a dev can set it up we can gather around 20 players to set the ratings.

Kaya-vv:
## Summary

Adds a rating system (0-1500) to the square color drill. Rating is computed from accuracy (65%) and response speed (35%), with a minimum of 20 questions required. The server recomputes the rating from session stats and persists the user's best-ever rating via a conditional DynamoDB update.

Frontend changes include a two-step Ready/GO start flow, a countdown showing questions remaining for a rating, and personal best display on both the Ready and Complete screens.

## Related Issues

Fixes #2149

## Type of change

## Testing

## Demo

**Two-step Ready/GO flow (second click starts the timer):**

Go btn

**Session complete with rating display:**

Rating

### DIFF
--- a/backend/puzzleService/squareColor.ts
+++ b/backend/puzzleService/squareColor.ts
@@ -1,15 +1,40 @@
-import { PutItemCommand } from '@aws-sdk/client-dynamodb';
+import { ConditionalCheckFailedException, PutItemCommand } from '@aws-sdk/client-dynamodb';
 import { marshall } from '@aws-sdk/util-dynamodb';
 import {
     SquareColorSessionResult,
+    SubmitSquareColorSessionResponse,
     submitSquareColorSessionSchema,
 } from '@jackstenglein/chess-dojo-common/src/squareColors/api';
+import {
+    computeSquareColorRating,
+    MIN_QUESTIONS_FOR_RATING,
+} from '@jackstenglein/chess-dojo-common/src/squareColors/rating';
 import { APIGatewayProxyHandlerV2 } from 'aws-lambda';
-import { errToApiGatewayProxyResultV2, parseBody, requireUserInfo, success } from '../directoryService/api';
-import { dynamo } from '../directoryService/database';
+import {
+    errToApiGatewayProxyResultV2,
+    parseBody,
+    requireUserInfo,
+    success,
+} from '../directoryService/api';
+import {
+    UpdateItemBuilder,
+    attributeNotExists,
+    dynamo,
+    lessThan,
+    or,
+} from '../directoryService/database';
 
 const squareColorResultsTable = `${process.env.stage}-square-color-results`;
+const usersTable = `${process.env.stage}-users`;
 
+/**
+ * Handles submission of a square color drill session. Saves the session result
+ * and, if enough questions were answered, computes a rating and conditionally
+ * updates the user's best rating.
+ *
+ * @param event - The API Gateway proxy event.
+ * @returns A response containing the computed rating if applicable, or an empty object.
+ */
 export const handler: APIGatewayProxyHandlerV2 = async (event) => {
     try {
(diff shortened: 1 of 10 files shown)


frontend/src/components/puzzles/squareColors/SquareColorDrillPage.tsx
```diff
@@ -325,6 +406,39 @@ function CompleteScreen({
...
+                            borderRadius: 2,
+                            backgroundColor: 'primary.main',
+                            color: 'primary.contrastText',
+                        }}
+                    >
+                        <Typography variant='overline' sx={{ opacity: 0.85 }}>
+                            Your Rating
+                        </Typography>
+                        <Typography variant='h3' sx={{ fontWeight: 'bold' }}>
+                            {summary.rating}
```
jalpp, line 429: we should show a up arrow green icon if their summary rating was greater than previous personal best, else down red arrow if it was opposite.

backend/puzzleService/squareColor.ts
```diff
@@ -22,14 +47,48 @@ export const handler: APIGatewayProxyHandlerV2 = async (event) => {
...
+                            attributeNotExists('squareColorRating'),
+                            lessThan('squareColorRating', response.rating),
+                        ),
+                    );
+                await builder.send();
+            } catch (err) {
+                if (!(err instanceof ConditionalCheckFailedException)) {
+                    throw err;
+                }
+                // Existing rating is already higher; no update needed.
```
jalpp: we don't need this comment if error occurs anyways updates won't be made

frontend/src/components/puzzles/squareColors/SquareColorDrillPage.tsx
```diff
@@ -373,6 +487,12 @@ function CompleteScreen({
     );
 }
 
+/**
+ * A single labeled stat row for the summary table.
+ *
+ * @param label - The human-readable label for the stat.
+ * @param value - The formatted value to display.
+ */
 function StatRow({ label, value }: { label: string; value: string }) {
```
jalpp: we should show reasonable icon for given stat and same green/red up/down arrow logic when stat is increasing/decreasing from previous attempt

jalpp (changes requested):

frontend/src/components/puzzles/squareColors/SquareColorDrillPage.tsx
Kaya-vv: @jalpp right now we only store the best-ever rating, not per-attempt stats. Delta arrows on each row would need atleast the previous attempt stored. Worth building that out, or is comparing against the personal best good enough?

jackstenglein: @Kaya-vv I think comparing against personal best is fine for now

jalpp (approved):
LGTM

PR #2176, opened 2026-03-30
prevent badge popup from re-triggering on every login

brianestraduh:
Summary
Fixes a race condition in BadgeCard where the badge popup (e.g. "Tactics Champion") would falsely re-trigger on every login. The root cause was that the earned-badge baseline was being set while requirements and timeline data were still loading, so when the full data arrived the component saw a "new" badge that was actually already earned.

The fix extracts badge detection into a pure detectNewBadge function and defers baseline initialization until both data sources have fully loaded (!isLoading()).

Related Issues
Fixes TECHGEN#932 (badge re-triggering component)

Bug report: "Upon login - I get a badge saying that my tactics are next level" — badge randomly pops up at least once a day.

Type of change

Bug fix (non-breaking change which fixes an issue)

New feature (non-breaking change which adds functionality)

Breaking change (fix or feature that would cause existing functionality to not work as expected)

Documentation update
Testing

New unit tests (vitest) were added

New playwright tests were added

It was not necessary to add tests due to {{reason}}
9 unit tests added in badgeHandler.test.ts covering:

Baseline not set while data is still loading (the race condition guard)
Baseline correctly set once data is fully loaded
Genuine new badges still detected during a session
End-to-end race condition scenario demonstrating the old broken behavior vs. the fix
Demo
No visual changes — the fix prevents a dialog from incorrectly appearing. Behavior is the same as before for genuinely new badges.


brianestraduh:
@jackstenglein the original defect was fixed regarding the missed progress
Screenshot_2026-03-24_at_4 20 46_PM
User Sent me this 
Screenshot_2026-03-29_at_6 14 13_PM
But user had this issue where this badge kept randomly appearing on log in. And this pull request addresses this issue. Let me know if you think this is needed.

jackstenglein (approved):

PR #2303, opened 2026-05-19
feat: redesign notification preference into table layout

### LINKED ISSUE #2301: Improve design of notification preference editor
Currently the notification preferences are organized by where the notification is delivered:

Image

As we add future notification channels (e.g. mobile), it will make more sense to sort these by notification type and then have a checkbox for each channel.

For example, a section titled `New Follower` with checkboxes for `Site`, `Email`, `Discord`

H4ROOON:
## Summary

refactored the notification preference ui into a table instead of grouped lists. switched it so notification types are rows and delivery methods (site, email, discord) are columns, which should make it easier to add stuff like mobile push later. kept all the existing preference mappings the same so current settings dont break. left new follower as site-only since email/discord support isnt there yet instead of showing fake checkboxes. also added a centered dash for unsupported cells so the table feels cleaner.

## Related Issues

closes #2301 

## Demo

image


jackstenglein:
Great, I really like this new UI!

PR #2312, opened 2026-05-21
refactor: per-section save/cancel buttons with unsaved changes dialog

### LINKED ISSUE #2306: make save/discard buttons sticky on settings pages
currently, the save and discard buttons on the profile settings pages are at the very top. if a user scrolls down on a long page (like notification preferences) and makes a change, they have to scroll all the way back to the top to save.

H4ROOON:
## Summary

refactored the settings page to use per-section save/cancel buttons. also added a navigation guard like the one on the game analysis board.

split the global update state into three isolated trackers for personal, ratings, and notifications. each section now has its own save/cancel buttons that only activate when something in that specific section changes. buttons are faded and non-clickable by default and light up when a change is detected.

upgraded getUpdate to deep compare objects with JSON.stringify since reference checks on ratings and notificationSettings were causing false positives and keeping buttons permanently lit. added a bypassGuard flag so saving and redirecting to /profile doesn't accidentally trip the nav guard before state clears. also fixed some default state mismatches like bio vs bio || '' so cancel properly zeroes out the form and re-disables the buttons.

## Related Issues

Closes #2306 

## Demo

image

Screenshot 2026-05-22 010043

### DIFF
--- a/frontend/src/app/(scoreboard)/profile/edit/ProfileEditorPage.tsx
+++ b/frontend/src/app/(scoreboard)/profile/edit/ProfileEditorPage.tsx
@@ -26,18 +26,23 @@ import NotInterestedIcon from '@mui/icons-material/NotInterested';
 import NotificationsIcon from '@mui/icons-material/Notifications';
 import SaveIcon from '@mui/icons-material/Save';
 import TimelineIcon from '@mui/icons-material/Timeline';
-import { LoadingButton } from '@mui/lab';
 import {
     Alert,
     Button,
     Card,
     CardContent,
     Container,
+    Dialog,
+    DialogActions,
+    DialogContent,
+    DialogContentText,
+    DialogTitle,
     Grid,
     Stack,
     Typography,
 } from '@mui/material';
-import React, { useState } from 'react';
+import { useNavigationGuard } from 'next-navigation-guard';
+import React, { useEffect, useState } from 'react';
 
 export const MAX_PROFILE_PICTURE_SIZE_MB = 9;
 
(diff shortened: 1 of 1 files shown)


frontend/src/app/(scoreboard)/profile/edit/ProfileEditorPage.tsx
```diff
@@ -140,70 +151,152 @@ export function ProfileEditorPage({ user }: { user: User }) {
...
-            timezoneOverride: timezone,
+            timezoneOverride:
+                timezone === DefaultTimezone && !user.timezoneOverride
+                    ? user.timezoneOverride
+                    : timezone,
+        },
+        profilePictureData,
+    );
+
+    const personalChangesMade =
```
jackstenglein: Couldn't this just be `const personalChangesMade = Boolean(personalUpdate)`? And the same below

frontend/src/app/(scoreboard)/profile/edit/ProfileEditorPage.tsx
```diff
@@ -348,85 +424,196 @@ export function ProfileEditorPage({ user }: { user: User }) {
...
+                            />
+                            <Stack direction='row' spacing={2} justifyContent='flex-end'>
                                 <LoadingButton
                                     variant='contained'
-                                    onClick={onSave}
+                                    onClick={personalChangesMade ? onSavePersonal : undefined}
                                     loading={request.status === RequestStatus.Loading}
-                                    disabled={!changesMade}
                                     startIcon={<SaveIcon />}
+                                    sx={{
```
jackstenglein: Let's just use `disabled={!personalChangesMade}` so that the disabled styling matches all the other buttons across the site. Then we also don't need to conditionally set onClick.

frontend/src/app/(scoreboard)/profile/edit/ProfileEditorPage.tsx
```diff
@@ -348,85 +424,196 @@ export function ProfileEditorPage({ user }: { user: User }) {
...
+                            </Stack>
+                        </Stack>
+
+                        <Stack spacing={2}>
+                            <NotificationSettingsEditor
+                                notificationSettings={notificationSettings}
+                                setNotificationSettings={setNotificationSettings}
+                            />
+                            <Stack direction='row' spacing={2} justifyContent='flex-end'>
+                                <LoadingButton
```
jackstenglein: LoadingButton is deprecated now in MUI, can we change this to just use Button?

jackstenglein (changes requested):

jackstenglein (approved):

Issue #2332, opened 2026-05-31
UX Improvements to Game Analysis


jackstenglein:
There are a number of UX improvements we would like to make to the Game Analysis page. Each of these could (and probably should) be separated into its own PR:

- [x] Allow opening the tabs on the analysis page in either window (either left or right of the board). Currently the PGN text is always on the right by itself and everything else is on the left. We want to allow putting any tab on either window or both. For example, a user might want the editor open on the right and the database explorer on the left. Or they might want two copies of the database explorer open so that they can view the dojo database on the left and the masters database on the right. I think there are a number of ways we could allow users to move the tabs between windows: right clicking on the tab to open a menu for moving, drag/drop, an option in the settings tab. I'm open to any/multiple of these, or others if whoever takes this issue has a good idea for it. Whatever is easy to implement/easy to use is best.
- [x] Add a setting that when viewing other user's games, variations are automatically left as comments. This avoids users having to right-click on each variation and clicking "save variation as comment"
- [x] Add a button that users can click which saves all unsaved variations as a comment. This could show up at the top of the PGN text when there are unsaved variations or somewhere in the comments tab or both (or somewhere else if you have a good idea).
- [x] Add a setting to display comment text inline in the PGN text. Currently we have a setting for displaying the suggested variations. We should add one for comments as well, and it should display similarly, where we have the user's avatar next to the comment text to indicate that they left the comment. Replies don't necessarily need to be included if it doesn't look good.
- [x] Allow users to delete moves from a suggested variation comment. Currently once a user puts a variation on the board, there is no way to remove it. We should only allow users to delete moves that they have suggested.
- [x] Add sounds for pieces moving/captures/check. Add a setting to turn off the sounds if desired.

If you comment on this and only want to take on one aspect of this ticket, please be clear about which part so others know what is still available to work on.

Kaya-vv:
I'll take the 3rd and 4th item:

- Add a button to save all unsaved variations as comments
- Add a setting to display comment text inline in the PGN text

If no one else picks up the remaining items after that, I may continue with more later.

H4ROOON:
ill take 2 and 5.

SushantVerma7969:
I've opened a PR for the sounds sub-task move, capture and check sounds with a toggle in settings.

Kaya-vv:
I'll take the remaining 1st item

PR #2335, opened 2026-06-01
feat(2332): add piece move/capture/check sounds with toggle setting

SushantVerma7969:
Picked up the sounds sub-task from #2332.

Adds audio feedback when moving pieces on the analysis board — 
different sounds for a regular move, a capture, and check. There's 
also a new checkbox in Viewer Settings ("Play sounds for piece moves") 
so users can turn it off if they don't want it.

The sounds are the standard ones from Lichess (MIT licensed), 
stored in `public/sounds/`.

Changes:
- `frontend/src/board/sounds/useBoardSound.ts` — new hook that picks 
  and plays the right sound based on the move
- `frontend/src/board/Board.tsx` — wired the hook into all 3 move paths
- `frontend/src/board/pgn/boardTools/underboard/settings/ViewerSettings.tsx` 
  — added the toggle under the existing Sounds section
- `frontend/public/sounds/` — Move.mp3, Capture.mp3, Check.mp3


frontend/public/sounds/Check.mp3
```diff
@@ -0,0 +1 @@
+../Silence.mp3
```
jackstenglein: This doesn't seem to be the right file

jackstenglein (changes requested):
Checks don't play a sound currently. Also, clicking on a move, clicking the arrows below the board and using the arrow keys on the keyboard do not play sounds. Only moving the piece manually on the board plays a sound right now. I think we should move triggering the sound into the reconcile function, which should handle all of the above cases.

SushantVerma7969:
Fixed both issues:

Check.mp3 — replaced the broken ../Silence.mp3 with the actual Lichess MP3 file (MIT licensed).
Sound triggering — moved playSound into the reconcile function. Sounds now fire for all move types: arrow keys, PGN move clicks, board arrows, and check — not just manual drag-and-drop.

PR #2339, opened 2026-06-02
feat: allow users to delete moves from suggested variations

H4ROOON:
## Summary

exposes the delete move menu options for variations if the current user is the variation author. mainline deletion rules still apply (delete before is properly disabled for sidelines).

## Related Issues

Related to #2332 (task5)

## Demo
image


jackstenglein:
@H4ROOON I merged this since it is much better already than what we currently have, but there is an issue that if you post the comment and then delete a move from the variation, that move reappears next time you refresh the page. We would need to edit the comment on delete to prevent that from happening. If you could open a follow-up PR for that, that would be great

PR #2397, opened 2026-06-05
fix: sync deleted suggested variations

### LINKED ISSUE #2396: bug: deleted moves in suggested variations reappear on page refresh

H4ROOON:
## Summary

fixed the ghost move bug on refresh
extracted the sync logic into a shared helper so both delete paths do the exact same thing.
added a while loop to walk up the tree so deep child deletions actually find the root and sync properly.
fixed a race condition by checking if the root survived *before* `chess.delete` mutates the pointers.

## Related Issues
Closes #2396

### DIFF
--- a/frontend/src/board/pgn/boardTools/underboard/DeletePrompt.tsx
+++ b/frontend/src/board/pgn/boardTools/underboard/DeletePrompt.tsx
@@ -1,9 +1,21 @@
+import { useApi } from '@/api/Api';
+import { GameApiContextType } from '@/api/gameApi';
+import { RequestSnackbar, useRequest } from '@/api/Request';
+import { useAuth } from '@/auth/Auth';
 import { useReconcile } from '@/board/Board';
+import useGame from '@/context/useGame';
+import { Game } from '@/database/game';
+import { User } from '@/database/user';
 import { Chess, Move } from '@jackstenglein/chess';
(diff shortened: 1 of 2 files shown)


frontend/src/board/pgn/boardTools/underboard/DeletePrompt.tsx
```diff
@@ -122,36 +173,83 @@ export interface DeletePromptProps {
...
+            return;
+        }
+
+        let rootToSync: Move | null = null;
+        let shouldSyncBackend = false;
+        let rootSurvived = false;
+
+        if (user && game) {
+            let current: Move | null = deleteAction.move;
+            while (current) {
```
jackstenglein: Is this loop necessary? A non-owner should only be able to delete the move if it is a suggested variation, so I'm having trouble coming up with a scenario where the if would be false on `current` but true on some move earlier in the tree

jackstenglein (changes requested):

frontend/src/board/pgn/boardTools/underboard/DeletePrompt.tsx
H4ROOON: i assumed the child node might be missing the dojoComment and added the loop as a guard. youre right, its not necessary, will update soon

jackstenglein (approved):

PR #2403, opened 2026-06-09
feat(#2332): Adds tabbed side panels to game analysis page

Kaya-vv:
## Summary

Adds tabbed side panels to the game analysis page. The analysis page can now show any analysis tab on either side of the board, including PGN Text, and supports two independent Explorer panels at the same time.

## Related Issues

Fixes item 1 of #2332

## Type of change

*   [ ] Documentation update

## Testing

## Demo

**Example of database explorer on both sides**
image


jackstenglein:
@Kaya-vv Thanks, this looks like a good start. I have a few feedback points before merging this:

Can we still have the banners above the tabs on the right? Currently you can only see them if you have the PGN text tab open. Also, the engine section and PGN text scrollbar look very strange now:

Screenshot 2026-06-11 at 1 59 10 PM

Let's also add the new right-side tabs to GamePage.tsx as well, so that once you save the game, you can still have tabs open on either side of the board. 

Finally, I think which tabs are on which side should be configurable in some way by the user and we shouldn't just have everything on both sides all the time, as that feels like it would be a bit overwhelming, especially to newer users.

Kaya-vv:
@jackstenglein thanks for the feedback. 

- Banners now stay above the right-side tabs.
- PGN Text/engine scrolling is cleaned up.
- Saved games now support right-side tabs too.
- Tab placement is configurable in Settings > Side Panels

image

PR #2419, opened 2026-06-16
fix: add color toggle to sparring bot

### LINKED ISSUE #2413: Cant choose "Play as" with maia sparring positions Tech_BUG#969
**📓 Steps to reproduce:** 
  > On the sparring positions page of the King's Indian Defense starter opening course, clicking on play bot defaults to "Play as white" with no option to change. This is not desirable because KID is an opening for black
🔨 Platform: chesdojo.club/Discord
📱 Device Type: Desktop
🌐 Browser Type: Chrome

👤 Author name: zaxili
[Discord URL](https://discord.com/channels/951958534113886238/1299913797045325957/1515817968037269606)

H4ROOON:
## Summary

added a color toggle to the play maia dialog for sparring positions. it defaults to the fen active turn but allows users to override it. replaced the hardcoded chip with a select dropdown to match the ui pattern of the practice menu.
tested across two different courses to confirm the fix isnt course-specific. in the KID starter course, sparring positions default to white, switched to black via the new dropdown and maia correctly responded as white first. in caro kann, positions default to black, switched to white and maia correctly responded as black first. confirmed both directions of the color override work and the bot always plays the opposite side correctly regardless of which way the user switches from the default.

## Related Issues

Closes #2413 

## Demo

image

### DIFF
--- a/frontend/src/components/playbot/PlayMaiaDialog.tsx
+++ b/frontend/src/components/playbot/PlayMaiaDialog.tsx
@@ -16,6 +16,7 @@ import {
     Stack,
     Typography,
 } from '@mui/material';
+import { useTranslations } from 'next-intl';
 import { useEffect, useState } from 'react';
 import { MAIA_RATINGS, MaiaRating } from './maiaengine';
 import { RATING_DESCRIPTIONS } from './playbot';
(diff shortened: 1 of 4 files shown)


frontend/src/components/playbot/PlayMaiaDialog.tsx
```diff
@@ -80,25 +85,45 @@ export function PlayMaiaDialog({
...
+                                        sx={{
+                                            width: 10,
+                                            height: 10,
+                                            borderRadius: '50%',
+                                            bgcolor: 'white',
+                                            border: '1px solid',
+                                            borderColor: 'divider',
+                                        }}
+                                    />
+                                    <span>Play as White</span>
```
jackstenglein: Can you add these labels to the new `frontend/messages` i18n files and then load them with the `useTranslations` hook instead? This should be simple enough that Google Translate can handle the German version

PR #2427, opened 2026-06-19
feat: add browser exit pop up when suggested variations are unsaved

### LINKED ISSUE #2426: Prevent exiting a tab with unsaved changes
With unsaved changes (suggested variations to a game), we should have a popup preventing one from exiting the browser

Image

H4ROOON:
## Summary

adds a native browser beforeunload listener directly to PgnBoard.tsx. it taps into the existing getUnsavedSuggestedVariationRoots logic.

now, if a user has "Auto Save Variations" turned off and attempts to close the tab with unsaved variations, the browser will natively block them with a "Leave site? Changes you made may not be saved" warning.

## Related Issues

Closes #2426 

## Demo

Screenshot 2026-06-19 150056

### DIFF
--- a/frontend/messages/de.json
+++ b/frontend/messages/de.json
@@ -1819,6 +1819,13 @@
             "saveAnalysis": "Analyse speichern",
             "analysisNotSaved": "Analyse nicht gespeichert"
         },
+        "unsavedNavigationGuard": {
+            "title": "Nicht gespeicherte Änderungen?",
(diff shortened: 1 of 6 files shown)


jackstenglein:
@H4ROOON The popup isn't triggered if the user performs a client-side navigation (for example, by switching to a different game using the folder tab). We should use the `useNavigationGuard` hook to handle client-side navigation as well. Take a look at AnalysisBoard.tsx for an example. Also, could you apply this to unsaved changes on an owned game as well? (IE: if the red cloud icon is below the board, then we should show a similar popup).

PR #2444, opened 2026-06-29
feat: open progress dialog when task is in progress

### LINKED ISSUE #2441: Clicking a task in progress should bring up Update Progress dialog
Instead of task info, clicking a task that's with any progress in it (whether in today's tasks or in full training plan) should pull up the Update Progress dialog.

H4ROOON:
## Summary

changes the click behavior in the training plan to default to the "Update Progress" dialog if the task has active progress (time or count). if the task has 0 progress, it defaults to the "Task Details" dialog.

includes a guard (isCurrentUser) so that viewing another user's profile still correctly defaults to "Task Details", preventing unauthorized progress updates. applied to Full, Weekly, and Daily plan views.

## Related Issues

Closes #2441


frontend/src/components/profile/trainingPlan/daily/DailyTrainingPlan.tsx
```diff
@@ -248,7 +248,14 @@ function DailyTrainingPlanItem({
             >
                 <CardActionArea
                     sx={{ flexGrow: 1 }}
-                    onClick={() => onOpenTask(task, TaskDialogView.Details)}
+                    onClick={() =>
+                        onOpenTask(
+                            task,
+                            isCurrentUser && timeWorkedMinutes > 0
```
jackstenglein: I think we should fetch the user's current count on this task and use that instead of `timeWorkedMinutes`. `timeWorkedMinutes` is only tracking the time they have worked in the current day, so this would still open the details the first time you click the task every day, even if it's something you already have progress on. Same thing in the WeeklyTrainingPlan

jackstenglein (changes requested):

H4ROOON:
updated with requested changes. also reused the variable for the progressCompleted below to avoid calculating twice per render

jackstenglein (approved):

PR #2503, opened 2026-07-15
feat: add draw offer ui to edit PGN in analysis board

### LINKED ISSUE #2495: Draw Offer TECHGEN#183
**💬 Content:** 
  > I recently learned that a draw offer is supposed to be recorded in notation as (=) after the move. (From what I understand, the parentheses distinguish it from = as an evaluation annotation.) Should we consider having a way to use this in our Dojo analysis board if it is part of standard notation? I'm not sure how that would translate into a PGN.

👤 Author name: dmlyum
[Discord URL](https://discord.com/channels/951958534113886238/1026952329804644372/1526082135767777511)

H4ROOON:
## Summary

added ui for marking a move as an offered draw in the pgn editor. the pgn text now displays `(=)` inline when `move.drawOffer` is set.

note: `@jackstenglein/chess`'s `renderPgn()` does not currently serialize `move.drawOffer` back into the pgn string. since save, export, and pdf download all rely on `renderPgn()`, the `(=)` annotation will display correctly in the ui but will not persist across saves or appear in exported files. flagged this for a follow-up pr or a fix in the core package. this pr only adds the frontend UI.

## Related Issues

Fixes #2495 

## Demo

image

### DIFF
--- a/frontend/messages/de.json
+++ b/frontend/messages/de.json
@@ -3519,6 +3519,7 @@
             "commentBefore": "Kommentar Davor",
             "commentAfter": "Kommentar danach",
             "viewMore": "Mehr anzeigen",
+            "offerDraw": "Remis anbieten",
             "nullMove": "Null",
             "makeMainLine": "Hauptlinie festlegen",
             "moveVariationUp": "Variation nach oben verschieben",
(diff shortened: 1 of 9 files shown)


jackstenglein:
@H4ROOON Is this notation standard? I haven't seen this anywhere else and we already have 2 NAGs for draw offer:

Screenshot 2026-07-15 at 7 03 01 PM

I would prefer to just continue using these unless there's a popular PGN editor that uses the other notation

H4ROOON:
@jackstenglein i looked it up and it is used in some older notations.
i agree, i havent seen it either. it makes sense to stick with current ones

Issue #2501, opened 2026-07-14
Fide rating suddenly reads as 0 Tech_BUG#243


jackstenglein:
**📓 Steps to reproduce:** 
  > The site used to read my fide profile correctly and identified my current fide elo. But loggin in recently it reads my fide rating as 0 and says I am in the wrong cohort.
🔨 Platform: chessdojo.club
📱 Device Type: Webpage on all devices
🌐 Browser Type: All browsers

👤 Author name: Elo
[Discord URL](https://discord.com/channels/951958534113886238/1299913797045325957/1526646037727547545)

Kaya-vv:
Ill look into this

PR #2543, opened 2026-07-21
feat(#2537): add monthly federation rating updates

### LINKED ISSUE #2537: Improvements to rating update pipeline
There are a few improvements we need to make to our pipeline for updating player ratings. For context, the current process is that a CloudWatch event triggers a lambda function every 24 hours. When triggering the lambda function, the event passes in a single cohort. The lambda then queries each user in that cohort, looks up their new ratings in all of their configured rating systems, and then writes the user back if their rating has changed. There are a couple problems with this approach (ordered in descending importance):

1. We have enough users now that some cohorts will hit the 15-minute lambda timeout and so some users don't actually get their ratings updated.
2. FIDE consistently returns a 500 error when we try to scrape their webpages.
3. We do not respect Chess.com 429 errors and do not retry, leading to many players not getting their ratings updated.
4. Chess.com and Lichess return 404s for some users, but then we query those users again the next day anyway. 
5. This was one of the earliest bits of code I wrote for the Dojo when I was a much worse programmer, and the code is somewhat low quality.

We would like to make the following updates:

- [x] Either use some sort of checkpointing system in lambda or switch from lambda to something like step functions to work around the timeouts.
- [x] Switch from scraping the FIDE website to downloading the latest rating list and fetching the ratings from there.
- [x] Respect chesscom 429 errors and use a retry system to fetch users that fail initially
- [x] Only update FIDE, USCF, etc. ratings monthly (for rating systems that only publish monthly rating lists)
- [x] Ideally we would mark usernames/IDs that return 404s and avoid processing those every time.

Kaya-vv:
## Summary

Replaces FIDE website scraping with a monthly rating-list import backed by DynamoDB. FIDE, USCF, ECF, KNSB, and ACF batch updates now run monthly because these federations publish ratings monthly or less frequently. Also adds failure monitoring and manual recovery support.

## Deployment Note

After deployment, manually invoke `downloadFideRatings` once to populate the new table. Until this initial import completes, FIDE lookups will return 404 responses.

## Related Issues

Fixes #2537 and #2501

## Type of change

* [ ] Documentation update

## Testing

* [x] New Go unit tests were added

### DIFF
--- a/backend/user/ratings/fide/main.go
+++ b/backend/user/ratings/fide/main.go
@@ -17,6 +17,7 @@ import (
 	"github.com/aws/aws-lambda-go/events"
 	"github.com/aws/aws-lambda-go/lambda"
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/dynamodb"
 	"github.com/aws/aws-sdk-go/service/dynamodb/dynamodbattribute"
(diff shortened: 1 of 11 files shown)


backend/user/ratings/fide/main.go
```diff
@@ -0,0 +1,375 @@
...
+	id := strings.TrimSpace(line[:p.idEnd])
+	if id == "" || strings.IndexFunc(id, func(r rune) bool { return r < '0' || r > '9' }) >= 0 {
+		return "", 0, false
+	}
+	field := strings.TrimSpace(line[p.ratingStart : p.ratingStart+srtngWidth])
+	if field == "" {
+		return id, 0, true
+	}
+	rating := 0
+	for _, r := range field {
```
jackstenglein: Why not use `strconv.Atoi`?

Kaya-vv: No particular reason, I replaced it.

